Issue #8, opened 2020-08-11
MBBSEmu - Parameterize appsettings.json file location


fletcherm:
The emulator currently makes assumptions about where to find `appsettings.json` file.

Let's parameterize it instead. That way we can be more explicit about where to find the file.

I'm thinking something like:
`MBBSEmu -S /foo/bar/appsettings.json`

I'm not married to the flag being `-S`. Whatever floats your boat.

fletcherm:
Ok, if I'm reading this correctly, I think what needs to be done is:

* move the ILogger _logger = ServiceResolver.GetService<ILogger> line down to get the logger out after the command line options are parsed
* Move the config = ServiceResolver.GetService<IConfiguration> to also be below the command line parsing
* add a case statement for pulling out a -S option for the appsettings.json
* somehow pass the above into the ServiceResolver when everything gets built so that it can honor the appsettings.json  file location

I /think/ that's what has the least impact here

fletcherm:
Notes from Discord conversation with @enusbaum 

* you can define the two variables with .GetService() calls below the command line switches
* just define ILogger where it is and assign it after the command line parsing
* you'll probably need to change the pattern in ServiceResolver though
* thats where it'll get hairy
* But you could just leave the static constructor empty
* and make a new Initalize() method
* and call that in program.cs to setup the Service Collection

fletcherm:
Should be all taken care of by https://github.com/enusbaum/MBBSEmu/pull/13

Issue #1, opened 2020-08-09
MBBSEmu can't load appsettings.json unless present in the cwd


fletcherm:
MBBSEmu exits with an error when `appsettings.json` is not in the current working directory.

I would expect MBBSEmu to, by default, look for the file to be in the same dir as it, regardless of what my cwd is.

```
λ MBBSEmu\bin\Release\netcoreapp3.1\MBBSEmu.exe
Critical Exception has occured:
System.TypeInitializationException: The type initializer for 'MBBSEmu.DependencyInjection.ServiceResolver' threw an exception.
 ---> System.IO.FileNotFoundException: Unable to locate appsettings.json file. Please ensure the file is in the same directory as the MBBSEmu executable file.
   at MBBSEmu.DependencyInjection.ServiceResolver.LoadAppSettings() in C:\git\MBBSEmu\MBBSEmu\DependencyInjection\ServiceResolver.cs:line 76
   at MBBSEmu.DependencyInjection.ServiceResolver..cctor() in C:\git\MBBSEmu\MBBSEmu\DependencyInjection\ServiceResolver.cs:line 31
   --- End of inner exception stack trace ---
   at MBBSEmu.Program.Main(String[] args) in C:\git\MBBSEmu\MBBSEmu\Program.cs:line 34
```

By contrast, MBBSEmu runs fine when it and `appsettings.json` are in the cwd

```
C:\git\MBBSEmu\MBBSEmu\bin\Release\netcoreapp3.1 (master -> origin)
λ .\MBBSEmu.exe
MBBSEmu - A MajorBBS/Worldgroup Emulator
By: Eric Nusbaum @ The MajorBBS Emulation Project
https://www.mbbsemu.com
https://github.com/enusbaum/MBBSEmu
-----------------------------------------------------------
Command Line Options:
(rest of output elided for brevity)
```

PR #13, opened 2020-08-13
Parameterize emulator config

fletcherm:
This addresses #8 

I decided not to try adding tests for this due to the static nature of things. IMO breaking this up a bit when I go to improve config file loading in general (for #1, and Paladine's currently open pull request). Then tests will be significantly easier to add.

Eager for someone else to manually test it and see where it does & doesn't hold up :)


MBBSEmu/Program.cs
```diff
@@ -96,6 +118,18 @@ static void Main(string[] args)
                     }
                 }
 
+                if (sSettingsFile == null)
```
fletcherm, line 121: @enusbaum Question about this `if` block and how C# default parameters work.

Can all this be condensed down to simply `ServiceResolver.Create(sSettingsFile)`?

I'm hoping that works, but my fear is that there's some subtle difference between the programmer explicitly setting `null`, vs. the variable implicitly being `null`, and how the default parameter in the `Create` function behaves.

Please tell me I'm overthinking this and we can delete some code here...

enusbaum: You could define default file name in this class as a constant and do something like `ServiceResolver.Create(sSettingsFile ?? DefaultFile)` (?? being an if-null-then) :)

fletcherm: Thanks - in learning more about `??` I also learned about `??=`. Ended up using that as IMO it removed some subtleties hiding in the syntax.

fletcherm:
@enusbaum Looks for your feedback here - the latest build failed.

First "Details" link for "Build #43 failed" gives me an error message that is meaningless to me

Second "Details" link gives me a 401

Are you able to make more sense of this?

PR #16, opened 2020-08-14
Change most of the socket code to async

paladine:
Socket accepting is now fully async. TelnetSession has async receiving implemented, but there's still the Sender thread that seems to only be used for the invoke echo stuff? We might want to try to revisit removing that. Would be nice to have no threads in TelnetSession at all. 

Improved the error handling.

RLogin hasn't been updated yet as I figured I'd get Telnet working smoothly first, then port everything.


paladine:
I screwed up the branches, let me redo it

PR #17, opened 2020-08-14
Update most socket code from synchronous -> asynchronous

paladine:
Socket accepting is now fully async. TelnetSession has async receiving implemented, but there's still the Sender thread that seems to only be used for the invoke echo stuff? We might want to try to revisit removing that. Would be nice to have no threads in TelnetSession at all.

Improved the error handling.

RLogin hasn't been updated yet as I figured I'd get Telnet working smoothly first, then port everything.

### DIFF
--- a/MBBSEmu/Server/Socket/SocketServer.cs
+++ b/MBBSEmu/Server/Socket/SocketServer.cs
@@ -1,4 +1,4 @@
-﻿using MBBSEmu.HostProcess;
+using MBBSEmu.HostProcess;
 using MBBSEmu.Session;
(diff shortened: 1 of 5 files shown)


MBBSEmu/Server/Socket/SocketServer.cs
```diff
@@ -1,4 +1,4 @@
-﻿using MBBSEmu.HostProcess;
+using MBBSEmu.HostProcess;
```
enusbaum, line 1: We can remove this import as well as the _host variable/ctor if it's no longer used.

PR #21, opened 2020-08-16
Implement a Ctrl-C handler and modify the server to exit cleanly.

paladine:
Has the added benefit that logging off with 'x' actually disconnects
the Telnet connection.

One ctrl-c will trigger shutdown and hopefully stop the server. A second
ctrl-c will forcibly kill the server if something doesn't shut down
cleanly.

Calls the `finrou` routine on all modules. Session logoffs also call
the `huprou` routine, though we might want to see if it's more natural
to call `stsrou` instead.

### DIFF
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -1,4 +1,4 @@
-﻿using MBBSEmu.CPU;
+using MBBSEmu.CPU;
 using MBBSEmu.Disassembler.Artifacts;
 using MBBSEmu.HostProcess.ExportedModules;
 using MBBSEmu.HostProcess.HostRoutines;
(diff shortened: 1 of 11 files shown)


MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -212,6 +212,37 @@ private void WorkerThread()
...
+            // kill all active sessions
+            foreach (SessionBase session in _channelDictionary.Values) {
+                session.Stop();
+            }
+
+            // let modules clean themselves up
+            callModuleRoutine("finrou");
+        }
+
+        private void callModuleRoutine(string routine, ushort channel = ushort.MaxValue) {
```
enusbaum, line 231: PascalCase for method names

paladine: Done

Issue #10, opened 2020-08-12
MBBSEmu - Global Command Processor


enusbaum:
Modules register global commands using `void globalcmd(int (*rouptr)());`.

MBBSEmu supports `globalcmd()` and the registered pointers are saved to `Module.GlobalCommandHandlers`, but the command handlers are not currently called in the main worker thread/event loop in MbbsHost.cs.

We need to enhance the worker thread/event loop in MbbsHost.cs to call the registered GlobalCommandHandlers registered for each module any time a user submits input (Status == 3).

https://github.com/enusbaum/MBBSEmu/blob/master/MBBSEmu/HostProcess/MbbsHost.cs#L114

Additional documentation for this feature is available in the [MajorBBS Developers Guide](https://download.mbbsemu.com/docs/MajorBBS_Developers_Guide_Rev_6.2_Jan_1994.pdf)

PR #25, opened 2020-08-18
Issue 10

enusbaum:
Fixes https://github.com/enusbaum/MBBSEmu/issues/10

Implemented handling of two types of Global Commands: 

1. System Global Routines using IGlobalRoutine
2. Module - by invoking Global Routine is one registered using `globalcmd()`

### DIFF
--- a/MBBSEmu/HostProcess/GlobalRoutines/IGlobalRoutine.cs
+++ b/MBBSEmu/HostProcess/GlobalRoutines/IGlobalRoutine.cs
@@ -1,0 +1,15 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using MBBSEmu.Memory;
+using MBBSEmu.Module;
+using MBBSEmu.Session;
+
+namespace MBBSEmu.HostProcess.GlobalRoutines
+{
+    public interface IGlobalRoutine
+    {
+        bool ProcessCommand(byte[] command, ushort channelNumber, PointerDictionary<SessionBase> sessions,
+            Dictionary<string, MbbsModule> modules);
+    }
+}
--- a/MBBSEmu/HostProcess/GlobalRoutines/IGlobalRoutine.cs
+++ b/MBBSEmu/HostProcess/GlobalRoutines/IGlobalRoutine.cs
@@ -1,14 +1,20 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Text;
-using MBBSEmu.Memory;
+﻿using MBBSEmu.Memory;
 using MBBSEmu.Module;
 using MBBSEmu.Session;
+using System.Collections.Generic;
 
 namespace MBBSEmu.HostProcess.GlobalRoutines
 {
     public interface IGlobalRoutine
     {
+        /// <summary>
+        ///     Processes a Global Command and returns TRUE if the command was handled
+        /// </summary>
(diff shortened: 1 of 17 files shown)


MBBSEmu/HostProcess/GlobalRoutines/IGlobalRoutine.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Text;
+using MBBSEmu.Memory;
+using MBBSEmu.Module;
+using MBBSEmu.Session;
+
+namespace MBBSEmu.HostProcess.GlobalRoutines
+{
+    public interface IGlobalRoutine
+    {
+        bool ProcessCommand(byte[] command, ushort channelNumber, PointerDictionary<SessionBase> sessions,
```
paladine, line 12: can you add some docs describing the return value at least?

MBBSEmu/HostProcess/GlobalRoutines/UsersOnlineGlobal.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    public class UsersOnlineGlobal : IGlobalRoutine
+    {
+        public bool ProcessCommand(byte[] command, ushort channelNumber, PointerDictionary<SessionBase> sessions, Dictionary<string, MbbsModule> modules)
+        {
+            var commandString = Encoding.ASCII.GetString(command).TrimEnd('\0');
+
+            if (commandString == "/#")
+            {
+                sessions[channelNumber].SendToClient("|RESET|\r\n|B||GREEN|LINE  USER-ID                    ....... OPTION SELECTED|RESET|\r\n".EncodeToANSIArray());
```
paladine, line 20: essentially a constant, would be a little performance improvement to store this as a static somewhere so it doesn't have to be parsed into ANSI each time

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -148,6 +157,41 @@ private void WorkerThread()
                     //Process a single incoming byte from the client session
                     session.ProcessDataFromClient();
 
+                    //Global Command Handler
+                    if (session.Status == 3 && DoGlobalsAttribute.Get(session.SessionState))
```
paladine, line 161: future todo, a descriptive enum instead of 3 would go a long way :)

MBBSEmu/HostProcess/GlobalRoutines/IGlobalRoutine.cs
enusbaum: Added xml-doc for this method

MBBSEmu/HostProcess/GlobalRoutines/UsersOnlineGlobal.cs
enusbaum: The ANSIStringExtensions helpers just help with giving the ANSI formatting a more readable form vs. "\0[39;49m" for colors -- even if it were static.

MBBSEmu/HostProcess/MbbsHost.cs
enusbaum: Agreed -- we should probably create an EnumChannelStatus to track the different codes and give some documentation on their meanings

PR #65, opened 2020-08-26
Properly call lonrou for Rlogin sessions

### LINKED ISSUE #56: MBBSEmu - Logon Routine Required
**Describe the bug**
When a user logs into the system, the function registered to `lonrou()` for each module is invoked. 

You're currently able to disable this functionality via the appconfig.json option `Module.DoLoginRoutine`.

Additionally, `lonrou()` is not being called for Rlogin clients

The problem with the logic is that modules assume `lonrou()` will be invoked with each login of a channel. Some modules (such as Tele-Arena) use `lonrou()` as an opportunity to load a player record.

**Proposed Solution**

1. Add a `bool` property on session base for `OutputEnabled` to specify if data is to be sent to the channel
2. In `SessionBase.SendToClient`, do nothing if `OutputEnabled` is set to `TRUE`
3. For Rlogin clients, on initial connect set `OutputEnabled` to `TRUE`
4. If the Rlogin client is connecting to the main port (for main menu), run all `lonrou()` routines as they would for a standard Telnet User
5. If the Rlogin client is connecting directly into a module, run the `lonrou()` for that module before the user is put into the module
6. Set `OutputEnabled` to `FALSE` after `lonrou()` has been executed for Rlogin CLients

Additionally ....

1. If appconfig.json setting `Module.DoLoginRoutine` is `TRUE`, set `OutputEnabled` to `FALSE` for all clients connecting via Telnet and set back to `TRUE` after `lonrou()` routines have executed.

paladine:
Resolves #56


paladine:
Was bored so added rlogin protocol parsing at the byte level

MBBSEmu/Session/Rlogin/RloginSession.cs
```diff
@@ -35,39 +38,70 @@ public RloginSession(IMbbsHost host, Socket rloginConnection, string moduleIdent
...
+            {
+                return (clientData, bytesReceived);
+            }
+
+            for (int i = 0; i < bytesReceived; ++i)
+            {
+                if (ProcessIncomingByte(clientData[i]))
+                {
+                    // return whatever data we may have left in the packet as client data
+                    int remaining = bytesReceived - i - 1;
```
enusbaum, line 100: var

MBBSEmu/Session/Rlogin/RloginSession.cs
```diff
@@ -35,39 +38,70 @@ public RloginSession(IMbbsHost host, Socket rloginConnection, string moduleIdent
...
+        }
+
+        protected override (byte[], int) ProcessIncomingClientData(byte[] clientData, int bytesReceived)
+        {
+            if (SessionState != EnumSessionState.Negotiating)
+            {
+                return (clientData, bytesReceived);
+            }
+
+            for (int i = 0; i < bytesReceived; ++i)
```
enusbaum, line 95: var

MBBSEmu/Session/SessionBase.cs
```diff
@@ -205,7 +210,11 @@ public string Username
         /// <param name="dataToSend"></param>
         public bool SendToClientAsync(byte[] dataToSend)
         {
-            return DataToClient.TryAdd(dataToSend.Where(c => shouldSendToClient(c)).ToArray());
+            if (OutputEnabled)
```
enusbaum, line 213: This needs to be added to SendToClient() as well

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -436,6 +440,26 @@ private void ExitModule(SessionBase session)
...
+        ///     Invokes routine registered as LONROU during MAJORBBS->REGISTER_MODULE() call on only
+        ///     the selected Rlogin module, specified by ModuleIdentifier.
+        ///
+        ///     Executes on the given channel once after a user successfully logs in.
+        /// </summary>
+        private void ProcessLONROU_FromRlogin(SessionBase session)
+        {
+            bool.TryParse(_configuration["Module.DoLoginRoutine"], out var doLoginRoutine);
+
+            if (doLoginRoutine)
```
enusbaum, line 453: `OutputEnabled = doLoginRoutine`, this config setting should only impact the output of the login routine not if it should be run. We should run login routines no matter what (as some modules require it), but give the user the option to suppress the spam.

We need to set the `OutputEnabled = doLoginRoutine` logic in the current `ProcessLONROU` as well, to always run but suppress spam if config is set to true.

Also -- is there a risk of running the login routine multiple times?

MBBSEmu/Session/SessionBase.cs
paladine: Good call

MBBSEmu/HostProcess/MbbsHost.cs
paladine: I think we want this always false in the RLogin case at least. I've moved the doLoginRoutine check into `ProcessLONROU` though

There should be no risk of running this routine twice. It's protected by the SessionState, so unless somehow the state gets reset back to LoginRoutines, it won't get called again.

PR #68, opened 2020-08-28
Added a Console Session Handler

enusbaum:
Implemented a Session Handler for a Console session. This would let a person play locally w/o having to Telnet or Rlogin. 

ANSI is handled by setting the console to Unicode and converting the extended ASCII codes to their Unicode counterparts. This allows both ANSI formatting and the ANSI characters to be displayed on the console.


MBBSEmu/Session/LocalConsole/LocalConsoleSession.cs
```diff
@@ -0,0 +1,95 @@
...
+
+                //Extended ASCII Characters are converted to their Unicode Counterparts
+                Console.Write(Encoding.Unicode.GetString(BitConverter.GetBytes(_extendedASCIItoUnicode[c])));
+            }
+        }
+
+        public override void Stop()
+        {
+            _consoleInputThreadIsRunning = false;
+            Environment.Exit(0);
```
paladine, line 92: this will prevent modules from shutting down cleanly no?

enusbaum: Let me see if I can invoke the Ctrl-C behavior...

OR we could always send them back to the login, so it's always running on the console until they Ctrl-C. This way someone could run MBBSEmu in a screen and just jump to the screen when they want to play vs. telnetting.

enusbaum: Modified the method to Call `Stop()` on the `MbbsHost` before exiting, this will give modules a chance to run their shutdown routines.

PR #78, opened 2020-08-31
Fixes for T-LORD (RTSLORD) and String Handling

### LINKED ISSUE #49: (RTSLORD) - T-LoRD Issues
1) Selecting Converse with patrons from the Inn locks up but does not crash MBBSEmu. After this occurs MBBSEmu still accepts connections but does not negotiate IAC or do anything other than log "HostProcess.MbbsHost.AddSession Session x.x.x.x:yyyyy added to incoming queue" until it is restarted.

2) There is no way to exit live chat (Jabber with the locals) or view the help menu. Chat doesn't work and randomly a new chat message clears the screen:

image

3) Write mail command cannot find any user specified.

4) List warriors has odd characters in the Mastered column:

image

5) Random characters such as shown in the list warriors screen shot are output during play sometimes they are accompanied by a screen clear.

Fighting, healing, buying/selling, bank deposit/withdraw, buying a room, flirting with violet/hearing seth all work. Longer play will be required but it seems close to being 100%.

enusbaum:
1. clrprf() writes null (0x0) to PRFBUF to ensure the string is null
2. strcpy() copies null (0x0) to the destination if the source is null
3. fgets() includes the last character read when encoutering \n or last byte in the file

Fixes #49

### DIFF
--- a/MBBSEmu.Tests/ExportedModules/Majorbbs/strcpy_Tests.cs
+++ b/MBBSEmu.Tests/ExportedModules/Majorbbs/strcpy_Tests.cs
@@ -13,6 +13,7 @@ namespace MBBSEmu.Tests.ExportedModules.Majorbbs
         [InlineData("TEST1\0", "TEST1\0")]
         [InlineData("TEST1\0TEST2\0", "TEST1\0")]
         [InlineData("\0", "\0")]
+        [InlineData("", "\0")]
         public void strcpy_Test(string sourceString, string expectedDestination)
         {
             //Reset State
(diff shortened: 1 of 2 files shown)


MBBSEmu.Tests/ExportedModules/Majorbbs/strcpy_Tests.cs
```diff
@@ -0,0 +1,35 @@
...
+namespace MBBSEmu.Tests.ExportedModules.Majorbbs
+{
+    public class strcpy_Tests : MajorbbsTestBase
+    {
+        private const int STRCPY_ORDINAL = 574;
+
+        [Theory]
+        [InlineData("TEST1\0", "TEST1\0")]
+        [InlineData("TEST1\0TEST2\0", "TEST1\0")]
+        [InlineData("\0", "\0")]
```
paladine, line 15: Where's the NULL test? :)

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -1725,6 +1726,7 @@ private void clrprf()
         {
             //Set prfptr to the base address of prfbuf
             Module.Memory.SetPointer("PRFPTR", Module.Memory.GetVariablePointer("PRFBUF"));
+            Module.Memory.SetByte(Module.Memory.GetVariablePointer("PRFBUF"), 0);
```
paladine, line 1729: maybe store the value from line 1728's GetVariablePointer and reuse here? Seems silly to repeat the same lookup you just did. 

As an future optimization, maybe it would make sense to cache all these common variable pointers so they don't have to be looked up

enusbaum: `GetVariablePointers` returns a reference from a Dictionary which is a hash table lookup and close to an O(1) operation -- essentially the cache 👍

MBBSEmu.Tests/ExportedModules/Majorbbs/strcpy_Tests.cs
enusbaum: 3rd test. The tests inits by writing `X` * 0xFF to the destination. Setting the source to null verifies that the destination is now null

paladine: That's not a NULL pointer though, it's a pointer to an array with \0 as the first element

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
paladine: grumble grumble O(1) is still less than O(2) ;-)

Can always be changed later

MBBSEmu.Tests/ExportedModules/Majorbbs/strcpy_Tests.cs
enusbaum: Checked in a test for this and added a little extra handling in `strcpy()` if the source is null

PR #88, opened 2020-09-02
Fixed parsin() to retain white spaces from the user input

### LINKED ISSUE #87: [MBBSEMU] rstrin doesn't restore extraneous spaces
**Module Information**
* Identifier MBBSEMU

rstrin replaces INPUT nulls with spaces. parsin parses the raw input and creates INPUT replacing the spaces with NULL

parsin consolidates multiple spaces into a single NULL but rstrin does not restore those multiple spaces as it should

**To Reproduce**
1. Go to MBBSEMU Echo mode
2. Type a string with multiple spaces in a row
3. It echos back the same string but without the multiple spaces

**Expected behavior**
The output should be exactly the same as what you input, spacing included

enusbaum:
Fixes #87 

parsin() retains the spaces (replaced with nulls), and margn/v is set to the non-null commands within the string.


paladine (changes requested):
Looks good, can you update Echo_Tests.cs to echo a command with extra spaces in it and have the module spit it back?

paladine (approved):

paladine (approved):

PR #90, opened 2020-09-03
Update MenuRoutines.cs to support custom Login ANSI file

tuday2:
Added Login.File setting to appsettings.json to specify a custom ANSI file to be used for prelogin display 

Checks to see if file exists and will display default if it cannot find what is specified -- if not specified it is passed as null and will fail the check and display default.

### DIFF
--- a/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
+++ b/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
@@ -182,7 +182,7 @@ namespace MBBSEmu.HostProcess.HostRoutines
         {
             EchoToClient(session, new byte[] { 0x1B, 0x5B, 0x32, 0x4A });
             EchoToClient(session, new byte[] { 0x1B, 0x5B, 0x48 });
-            string LoginFileName = _configuration["Welcome.File"];
+            string LoginFileName = _configuration["Login.File"];
             //Check if file is specified in appsettings.json and if it exists, else display default
             if (File.Exists(LoginFileName)) {
                 byte[] LoginFileContents = File.ReadAllBytes(LoginFileName);
--- a/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
+++ b/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
@@ -315,7 +315,7 @@ namespace MBBSEmu.HostProcess.HostRoutines
             if (inputValue == 'X')
             {
                 session.SessionState = EnumSessionState.ConfirmLogoffDisplay;
-                //Load File if specified in appsettings.json and display if it exists, else display default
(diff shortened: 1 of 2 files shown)


MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
```diff
@@ -178,7 +182,14 @@ private void WelcomeScreenDisplay(SessionBase session)
         {
             EchoToClient(session, new byte[] { 0x1B, 0x5B, 0x32, 0x4A });
             EchoToClient(session, new byte[] { 0x1B, 0x5B, 0x48 });
-            EchoToClient(session, _resourceManager.GetResource("MBBSEmu.Assets.login.ans").ToArray());
+            string LoginFileName = _configuration["Login.File"];
```
enusbaum, line 185: I think we should rename the key to `ANSI.Login` only because I think we'll have an opportunity to use it in a couple more places, like `ANSI.Singup` and perhaps `ANSI.Logoff` and it'd keep them alphabetically grouped.

enusbaum, line 185: var

MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
```diff
@@ -178,7 +182,14 @@ private void WelcomeScreenDisplay(SessionBase session)
...
+            //Check if file is specified in appsettings.json and if it exists, else display default
+            if (File.Exists(LoginFileName)) {
+                byte[] LoginFileContents = File.ReadAllBytes(LoginFileName);
+                EchoToClient(session, LoginFileContents.ToArray());
```
enusbaum, line 189: You can inline `File.ReadAllBytes()` here not having to declare the variable since it's only used the one time

enusbaum (changes requested):

tuday2:
@enusbaum  Addressed comments and added custom SignUp and Logoff ANSI files

MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
```diff
@@ -298,6 +315,11 @@ private void MainMenuInput(SessionBase session, Dictionary<string, MbbsModule> m
             if (inputValue == 'X')
             {
                 session.SessionState = EnumSessionState.ConfirmLogoffDisplay;
+                //Load File if specified in appsettings.json and display if it exists
+                var ANSILogoffFileName = _configuration["ANSI.Logoff"];
```
enusbaum, line 319: We should move this to after a user confirms logging of, replacing the default text `Ok, thanks for calling, etc.` in `LoggingOffDisplay()`

enusbaum (changes requested):

enusbaum (approved):

PR #92, opened 2020-09-04
Write a bunch of libc tests, and reimplemented a bunch while I was at it.

paladine:


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -4327,29 +4312,10 @@ private void stpbtvl()
...
-                if (string1[i] == string2[i]) continue;
-
-                //1 < 2 == -1 (0xFFFF)
-                //1 > 2 == 1 (0x0001)
-                Registers.AX = (ushort)(string1[i] < string2[i] ? 0xFFFF : 1);
-                return;
-            }
-
-            Registers.AX = 0;
+            Registers.AX = (ushort)(short)String.Compare(string1, string2);
```
enusbaum, line 4318: `string` vs `String` alias
Unnecessary cast to `(short`) before `(ushort)`. `int` -1 will cast to `ushort` 0xFFFF

image

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -4893,47 +4859,52 @@ private void gabbtv()
...
 
-            if (!Module.Memory.TryGetVariablePointer("STROK-RESULT", out var resultPointer))
-                resultPointer = Module.Memory.AllocateVariable("STROK-RESULT", 0x400);
-
-            if (!Module.Memory.TryGetVariablePointer("STROK-ORIGINAL", out var originalPointer))
-                originalPointer = Module.Memory.AllocateVariable("STROK-ORIGINAL", 0x400);
-
-            if (!Module.Memory.TryGetVariablePointer("STROK-ORDINAL", out var ordinalPointer))
-                ordinalPointer = Module.Memory.AllocateVariable("STROK-ORDINAL", 0x2);
+            var workPointerPointer = Module.Memory.GetOrAllocateVariablePointer("STRTOK-WRK", 4);
```
enusbaum, line 4865: I started using the `IntPtr16.Size` const on declarations just to denote what'll be saved there

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -5056,50 +5027,10 @@ private void fgetc()
...
-                if (string1[i] == string2[i] - 32) continue;
-
-                //1 < 2 == -1 (0xFFFF)
-                //1 > 2 == 1 (0x0001)
-                Registers.AX = (ushort)(string1[i] < string2[i] ? 0xFFFF : 1);
-                return;
-            }
-
-            Registers.AX = 0;
+            Registers.AX = (ushort)(short)String.Compare(string1, string2, /* ignoreCase= */ true);
```
enusbaum, line 5033: `string` vs `String` alias
`(short)` casting can be removed
You can use parameter naming in C#, so `ignoreCase: true` would work for the 3rd parameter vs. the inline comment

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -5719,20 +5652,11 @@ private void memcmp()
...
-                if (ptr1Data[i] == ptr2Data[i])
-                    continue;
-
-                if (ptr1Data[i] > ptr2Data[i])
-                    Registers.AX = 0x1;
-                else
-                    Registers.AX = 0xFFFF;
-
-                return;
+                Registers.AX = (ushort)(short)(ptr1Data[i] - ptr2Data[i]);
```
enusbaum, line 5658: `(short)` casting can be removed

enusbaum (changes requested):

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
paladine: Cool, using magic numbers like true/false in method calls can be confusing if you're not sure what the parameter is for, but parameter naming takes that ambiguity away

PR #95, opened 2020-09-06
Add PageUserGlobal for paging functionality

tuday2:
I forgot to merge so had to revert, merge, and redo -- let me know if I did something wrong.

Add InstantMessageGlobal to enable paging functionality

Stayed true to look/feel from MajorBBS - may look at globals packages to improve look/feel or perhaps add a few more options

Submitting for initial feedback before I continue @enusbaum


MBBSEmu/HostProcess/GlobalRoutines/InstantMessageGlobal.cs
```diff
@@ -0,0 +1,51 @@
...
+namespace MBBSEmu.HostProcess.GlobalRoutines
+{
+    public class InstantMessageGlobal : IGlobalRoutine
+    {
+        public bool ProcessCommand(byte[] command, ushort channelNumber, PointerDictionary<SessionBase> sessions, Dictionary<string, MbbsModule> modules)
+        {
+            var commandString = Encoding.ASCII.GetString(command).TrimEnd('\0');
+            
+            //Check minimum length of a /P command (one character username/one character message = 6) and make sure there are at least 2 spaces
+            if (commandString.Length > 6 && commandString.Substring(0,2).ToUpper() == "/P" && Regex.Matches(commandString, " ").Count > 1)
```
enusbaum, line 19: You can do `commandString.StartsWith("/P", StringComparison.InvariantCultureIgnoreCase)` vs. the substring
also you can count the occurrences of space in a string using LINQ by doing `commandString.Count(c=> c == ' ')`

MBBSEmu/HostProcess/GlobalRoutines/InstantMessageGlobal.cs
```diff
@@ -0,0 +1,51 @@
...
+            {
+                //Determine location of first 2 spaces to separate username and message
+                var instantMessageSpace1 = commandString.IndexOf(' ');
```
enusbaum, line 22: It might be easier to `commandString.Split(' ')` to split up the individual words, getting `[1]` for the username then doing a `string.Join` on the remaining elements to get the message.

MBBSEmu/HostProcess/GlobalRoutines/InstantMessageGlobal.cs
```diff
@@ -0,0 +1,51 @@
...
+                if (sessions.Values.Count(u => u.Username.StartsWith(instantMessageTargetUser)) == 1)
+                {
+                    foreach (var s in sessions.Values)
+                    {
+                        var tempUserName = s.Username;
+                        if (tempUserName == instantMessageTargetUser || tempUserName.StartsWith(instantMessageTargetUser))
+                        {
+                            sessions[channelNumber].SendToClient($"|RESET|\r\n|B||YELLOW|... Paging {tempUserName} ...|RESET|\r\n".EncodeToANSIArray());
+                            // TO DO: Need to pull module name, currently says "InModule"
+                            sessions[s.Channel].SendToClient($"|RESET|\r\n|B||YELLOW|{instantMessageSourceUser} is paging you from {sessions[channelNumber].SessionState}: {instantMessageText}|RESET|\r\n".EncodeToANSIArray());
```
enusbaum, line 40: I have a routine to get the module/state name in `UsersOnlineGlobal.cs` but it's a little long. We might want to make a Globals base class that has this common functionality in it.

MBBSEmu/HostProcess/GlobalRoutines/InstantMessageGlobal.cs
```diff
@@ -0,0 +1,51 @@
...
+                var instantMessageSpace1 = commandString.IndexOf(' ');
+                var instantMessageSpace2 = commandString.IndexOf(' ', instantMessageSpace1 + 1);
+
+                //Define source username, target username, and message
+                var instantMessageSourceUser = sessions[channelNumber].Username;
+                var instantMessageTargetUser = commandString.Substring(3, instantMessageSpace2-3);
+                var instantMessageText = commandString.Substring(instantMessageSpace2+1);
+
+                //Check to see if the target user matches or matches part of any logged in users
+                if (sessions.Values.Count(u => u.Username.StartsWith(instantMessageTargetUser)) == 1)
```
enusbaum, line 31: You can change this from a `Count` to `Where` so you're retrieving the list of users that match (no need for the nested foreach). 

If it's > 1, display the error and return. 

Otherwise, get the `[0]` value as it's the target user then send

enusbaum (changes requested):

tuday2:
### **For Reference:**

**/P Command Options**
MajorBBS SlashP Input

**Result of a /P command:**
MajorBBS SlashP Output

MBBSEmu/HostProcess/GlobalRoutines/InstantMessageGlobal.cs
tuday2: I see that -- in this case we really only need it for the inModule isntance -- will review how we can break it out and use for both, i do like the "Friendlier" names you gave things in that giant switch statement

tuday2: Simple 1 line detection for modules, will explore creating Globals base class

MBBSEmu/HostProcess/GlobalRoutines/InstantMessageGlobal.cs
```diff
@@ -0,0 +1,69 @@
...
+using System.Text;
+using MBBSEmu.Session.Enums;
+
+namespace MBBSEmu.HostProcess.GlobalRoutines
+{
+    public class InstantMessageGlobal : IGlobalRoutine
+    {
+        public bool ProcessCommand(byte[] command, ushort channelNumber, PointerDictionary<SessionBase> sessions, Dictionary<string, MbbsModule> modules)
+        {
+            var commandString = Encoding.ASCII.GetString(command).TrimEnd('\0');
```
enusbaum, line 17: We can do `if(command.Length < 6) return;` here to bail out quickly before converting the input. Also simplifies the below IF statement.

enusbaum (changes requested):

tuday2:
@enusbaum  should be good, cleaned up naming, removed nested IF, and addressed your comment on bailing out if <6.

MBBSEmu/HostProcess/GlobalRoutines/PageUserGlobal.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace MBBSEmu.HostProcess.GlobalRoutines
+{
+    public class PageUserGlobal : IGlobalRoutine
+    {
+        public bool ProcessCommand(byte[] command, ushort channelNumber, PointerDictionary<SessionBase> sessions, Dictionary<string, MbbsModule> modules)
+        {
+            var commandString = Encoding.ASCII.GetString(command).TrimEnd('\0');
+            
+            //Check minimum length for a /p command else exit
+            if(commandString.Length < 6) return false;
```
enusbaum, line 20: Move this before `commandString` is declared and have it evaluate the length of `command`. Because globals are checked on every user input, just want to make it as performant as possible.

enusbaum (changes requested):

MBBSEmu/HostProcess/GlobalRoutines/PageUserGlobal.cs
tuday2: Fixed @enusbaum

enusbaum (approved):

PR #98, opened 2020-09-09
Make Telnet processing packet agnostic.

paladine:
Reworks the telnet processing. Seems to work and doesn't send any duplicate negotiations thanks to the hashmap.


enusbaum (changes requested):
Tested it with Synchro and Putty and it works great!

enusbaum (approved):

Issue #118, opened 2020-09-12
[ELWGOLF] Electronic Fairways - Module crashes during gameplay


tuday2:
**Module Information**
* Identifier: ELWGOLF
* Module Name: Electronic Fairways
* Version: 2.0

**Describe the bug**

**To Reproduce**
1. Enter Module
2. (P)lay Game
3. Select club and strength

**Expected behavior**
Ability to play through a shot

**Screenshots**
![Screenshot 2020-09-12 081948](https://user-images.githubusercontent.com/27604078/92996391-e163a400-f4d0-11ea-963b-04ef882e2b22.JPG)

```
Unhandled exception. System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'Unsupported OpCode: Fdivp')
   at MBBSEmu.CPU.CpuCore.Tick()
   at MBBSEmu.HostProcess.ExecutionUnits.ExecutionUnit.Execute(IntPtr16 entryPoint, UInt16 channelNumber, Boolean simulateCallFar, Boolean bypassState, Queue`1 initialStackValues, UInt16 initialStackPointer)
   at MBBSEmu.Module.MbbsModule.Execute(IntPtr16 entryPoint, UInt16 channelNumber, Boolean simulateCallFar, Boolean bypassSetState, Queue`1 initialStackValues, UInt16 initialStackPointer)
   at MBBSEmu.HostProcess.MbbsHost.WorkerThread()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

**Software Information:**
- OS: Windows 2019 Server Standard
- Telnet Client: NetRunner

enusbaum:
`FDIVP` - Similar to `FADDP`. Divide ST(1) by ST(0), store result in ST(1), and pop the register stack effectively making the result stored in ST(1) now ST(0).

tuday2:
Getting this error now -- is this related or a new issue? @enusbaum @paladine 

Unhandled exception. System.Exception: Type Requested not UInt16
[...]

PR #119, opened 2020-09-12
x87 fixes

enusbaum:
Fixes and Unit Tests for x87 Opcodes. 

Verified x87 Operation in both Unit Tests and a Test Worldgroup Module that performed multiple Floating Point operations.

### DIFF
--- a/MBBSEmu.Tests/CPU/CpuTestBase.cs
+++ b/MBBSEmu.Tests/CPU/CpuTestBase.cs
@@ -2,10 +2,8 @@
 using MBBSEmu.CPU;
 using MBBSEmu.Memory;
 using System;
-using System.Collections.Generic;
 using System.IO;
-using System.Linq;
-using static Iced.Intel.AssemblerRegisters;
+
 namespace MBBSEmu.Tests.CPU
 {
     public abstract class CpuTestBase : TestBase
--- a/MBBSEmu.Tests/CPU/CpuTestBase.cs
+++ b/MBBSEmu.Tests/CPU/CpuTestBase.cs
@@ -3,7 +3,9 @@ using MBBSEmu.CPU;
 using MBBSEmu.Memory;
 using System;
 using System.Collections.Generic;
-
+using System.IO;
+using System.Linq;
+using static Iced.Intel.AssemblerRegisters;
 namespace MBBSEmu.Tests.CPU
 {
     public abstract class CpuTestBase : TestBase
@@ -29,14 +31,12 @@ namespace MBBSEmu.Tests.CPU
             mbbsEmuCpuRegisters.IP = 0;
         }
 
-        protected void CreateCodeSegment(IReadOnlyList<Instruction> instructions, ushort segmentOrdinal = 1)
+        protected void CreateCodeSegment(Assembler instructions, ushort segmentOrdinal = 1)
         {
-            var instructionList = new InstructionList(instructions.Count);
+            var stream = new MemoryStream();
+            instructions.Assemble(new StreamCodeWriter(stream), 0);
 
-            foreach(var i in instructions)
-                instructionList.Add(i);
-
-            CreateCodeSegment(instructionList, segmentOrdinal);
+            CreateCodeSegment(stream.ToArray(), segmentOrdinal);
         }
 
         protected void CreateCodeSegment(ReadOnlySpan<byte> byteCode, ushort segmentOrdinal = 1)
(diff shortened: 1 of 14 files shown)


MBBSEmu.Tests/CPU/CpuTestBase.cs
```diff
@@ -3,7 +3,9 @@
 using MBBSEmu.Memory;
 using System;
 using System.Collections.Generic;
-
+using System.IO;
+using System.Linq;
+using static Iced.Intel.AssemblerRegisters;
```
paladine, line 8: blank line between using statements and start of namespace?

MBBSEmu.Tests/CPU/FADDP_Tests.cs
```diff
@@ -0,0 +1,33 @@
...
+
+namespace MBBSEmu.Tests.CPU
+{
+    public class FADDP_Tests : CpuTestBase
+    {
+        [Theory]
+        [InlineData(0.5, 1.5)]
+        [InlineData(float.MaxValue, 1)]
+        [InlineData(float.MinValue, float.MaxValue)]
+        [InlineData(0, 0)]
```
paladine, line 13: any tests for `Single.NaN`? ;-) And double negatives?

MBBSEmu.Tests/CPU/FADD_Tests.cs
```diff
@@ -0,0 +1,36 @@
...
+        [Theory]
+        [InlineData(0.5, 1.5)]
+        [InlineData(float.MaxValue, 1)]
+        [InlineData(float.MinValue, float.MaxValue)]
+        [InlineData(0, 0)]
+        public void FADD_Test(float initialST0Value, float valueToAdd)
+        {
+            Reset();
+
+            CreateDataSegment(new ReadOnlySpan<byte>(), 2);
```
paladine, line 19: does this return the segment? (I thought I made that change).

I'd find it easier to understand if you stored the segment return value and used it in the calls to SetArray and DS = <val>

MBBSEmu.Tests/CPU/FDIV_Tests.cs
```diff
@@ -0,0 +1,35 @@
...
+using static Iced.Intel.AssemblerRegisters;
+
+namespace MBBSEmu.Tests.CPU
+{
+    public class FDIV_Tests : CpuTestBase
+    {
+        [Theory]
+        [InlineData(2, 2)]
+        [InlineData(1, .5)]
+        [InlineData(0, 0)]
```
paladine, line 13: divide by zero is a great test, doesn't the test bomb if you attempt it?

MBBSEmu/CPU/CPUCore.cs
```diff
@@ -2260,68 +2320,63 @@ private void Op_Call()
...
-            Registers.Fpu.PopStackTop();
-            var float1 = BitConverter.ToSingle(FpuStack[Registers.Fpu.GetStackTop()]);
-            var float2 = BitConverter.ToSingle(floatToMultiply);
+            var ST0 = FpuStack[Registers.Fpu.GetStackTop()];
+            var source = GetOperandValueFloat(_currentInstruction.Op0Kind, EnumOperandType.Source);
 
-            var result = float1 * float2;
-            FpuStack[Registers.Fpu.GetStackTop()] = BitConverter.GetBytes(result);
-            Registers.Fpu.PushStackTop();
+            var result = ST0 * source;
```
paladine, line 2357: do these need to be wrapped in unchecked?

paladine (changes requested):

MBBSEmu.Tests/CPU/FADD_Tests.cs
enusbaum: It doesn't -- it just creates an empty segment at the specified ordinal. You have the option to pass in the data to load the segment with initially, which in this case is just nothing.

MBBSEmu.Tests/CPU/FDIV_Tests.cs
enusbaum: Nope! Even unwrapped from the unchecked it just returns `NaN`

MBBSEmu/CPU/CPUCore.cs
enusbaum: You know, I had these there because I thought it wouldn't gracefully handle the overflow but it seems to handle it differently than integer types where it just casts it as NaN. I went ahead and removed it from the tests.

enusbaum:
Fixes #118

paladine (approved):

Issue #35, opened 2020-08-20
Selecting modules from main menu


dlightman311:
There seems to be an input handling issue in build 081920-8. If more than 9 modules are loaded attempts to access modules with an option number of greater than 9 seem to have the second digit truncated loading module 1.

Additionally after this occurs BladeMaster and Blox are left in a inaccessible state where only the "module menu" is displayed then returning to the main menu. Restarting MBBSEmu restores access until a module > 9 is selected again. 

Captures:

[first.txt](https://github.com/enusbaum/MBBSEmu/files/5103768/first.txt)
[second.txt](https://github.com/enusbaum/MBBSEmu/files/5103776/second.txt)

enusbaum:
What maniac would ever want to run more than 9 modules?! ;)

https://github.com/enusbaum/MBBSEmu/blob/master/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs#L284

fletcherm:
Started tinkering with this -- got an environment setup with a bunch of modules, reproduced the bug, and played around with what a fix might look like.

Here's a diff of what I did that got things working better
```
-            var inputValue = Encoding.ASCII.GetString(session.InputBuffer.ToArray()).ToUpper()[0];
+            var inputValue = Encoding.ASCII.GetString(session.InputBuffer.ToArray());

             //User is Logging Off
-            if (inputValue == 'X')
+            if (inputValue.ToUpper() == "X")
```

Crashes when I enter `-10` so that needs to be improved as well.

Next step: stash these changes, create some failing tests, then apply the fixes.

PR #120, opened 2020-09-13
Update MenuRoutines.cs

tuday2:
*Fixed Issue #35  to allow inputs > 1 digit
*Cleaned up If Statements from custom ANSI files PR
*Fixed crash caused by negative number input to menu

### DIFF
--- a/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
+++ b/MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
@@ -285,7 +285,7 @@ namespace MBBSEmu.HostProcess.HostRoutines
             var currentMenuItem = 0;
             foreach (var m in modules)
             {
-                EchoToClient(session, $"   |CYAN||B|{currentMenuItem}|YELLOW| ... {m.Value.ModuleDescription}\r\n".EncodeToANSIArray());
+                EchoToClient(session, $"   |CYAN||B|{currentMenuItem}|YELLOW| ... {m.Value.ModuleDescription}\r\n".PadRight(3,' ').EncodeToANSIArray());
                 currentMenuItem++;
             }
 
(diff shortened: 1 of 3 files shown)


MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
```diff
@@ -308,18 +306,29 @@ private void MainMenuInput(SessionBase session, Dictionary<string, MbbsModule> m
                 return;
             }
 
-            var inputValue = Encoding.ASCII.GetString(session.InputBuffer.ToArray()).ToUpper()[0];
+            var inputValue = Encoding.ASCII.GetString(session.InputBuffer.ToArray());
+            
+            //convert to uppercase, trim
+            var inputCommand = inputValue.ToUpper().TrimEnd('\0');
 
+            //Exit if command longer then 2 characters (up to 99 modules) or a negative number
```
paladine, line 314: s/then/than

MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
```diff
@@ -308,18 +306,29 @@ private void MainMenuInput(SessionBase session, Dictionary<string, MbbsModule> m
...
+            if (inputCommand.Length > 2 || inputCommand.Contains('-'))
```
paladine, line 315: This entire section might not be necessary. 

Leave the inputCommand == "X" logic, and then relying on `Int.TryParse` should be sufficient no? Probably have to add an additional check on selectedMenuItem >= 0

paladine (changes requested):

MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
tuday2: removed.

MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
tuday2: Added OR statement with >= 0 and removed extra check. Only outstanding item is you can type 00000000 and get the 0 menu item... not sure it matters but that's the only other thing that slips through

enusbaum: I'm fine with 00000 still resolving to 0

enusbaum:
Fixes #35

Issue #132, opened 2020-09-17
Nightly cleanup doesn't call module exit/reinit functions


paladine:
The nightly cleanup routine calls `mcurou` only. But it really should call `mcurou` followed by `finrou` and then `init_xxx` to restart it

tuday2:
address in PR #144

PR #144, opened 2020-09-20
Update MenuRoutines.cs (Cleanup variable) and address Nightly Cleanup issue  #132 and add Notification to Users

tuday2:
* Clean up old variable not needed anymore


tuday2:
Screenshot 2020-09-20 210437

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -977,10 +978,20 @@ private void DoNightlyCleanup()
...
+                _channelDictionary[c.Value.Channel].SendToClient($"|RESET|\r\n|B||RED|Nightly Cleanup Running -- Please log back on shortly|RESET|\r\n".EncodeToANSIArray());
+            
             // removes all sessions
             RemoveSessions(session => true);
 
             CallModuleRoutine("mcurou", module => _logger.Info($"Calling nightly cleanup routine on module {module.ModuleIdentifier}"));
+            CallModuleRoutine("finrou", module => _logger.Info($"Calling finish-up (sys-shutdown) routine on module {module.ModuleIdentifier}"));
+            foreach (var m in _modules)
+            {
+                Run(m.Value.ModuleIdentifier, m.Value.EntryPoints["_INIT_"], ushort.MaxValue);
```
enusbaum, line 992: We might need to add some additional checks on some things that happen at startup. One off the top of my head is text variable registration (`REGISTER_TEXTVAR`) and how MBBSEmu will happen if they're called again -- things like that we might need to make more resilient to do a Check/Add

enusbaum (changes requested):
Need to test this on a couple more modules that have more intricate startups.

MBBSEmu/HostProcess/MbbsHost.cs
tuday2: Testing with both variations (below) which fix crash with DIALCHAT. 

Trying a wide variety of modules and checking variables/module states before/after

**TryAdd:**
Module.TextVariables.TryAdd(Encoding.ASCII.GetString(textBytes), functionPointer);

**Update/Add:**
Module.TextVariables[Encoding.ASCII.GetString(textBytes)] = functionPointer;

enusbaum: We'll want to verify anywhere a Dictionary is defined in the ExportedModule or the Base Class, that it is using a safe add method. I think there are only a couple, but just need to verify.

enusbaum (changes requested):
Need to verify with any dictionary used in an exported module/exported module base class that it's using a safe add. MCV file might be another one.

tuday2:
Will check dictionary's....

Also, I did a compare against the initial load debug text and the "INIT" load debug text and looking at the differences, going to paste some screenshots and will be looking into the routines calling these.....

tuday2:
Memory Locations are different -- should this be running again?

Screenshot 2020-09-23 223101

tuday2:
Memory Locations are different -- should this be running again?

Screenshot 2020-09-23 223558222

tuday2:
"Re-Acclocate" variables -- definitely think I need to check into this

Screenshot 2020-09-23 223909333

tuday2:
For Reference

[DialChat.Cleanup.log](https://github.com/enusbaum/MBBSEmu/files/5273294/DialChat.Cleanup.log)
[DialChat.Load.log](https://github.com/enusbaum/MBBSEmu/files/5273296/DialChat.Load.log)

enusbaum:
From our Discussion, the cleanest strategy for this implementation will be to essentially replicate the same process that's in MajorBBS/Worldgroup where the module is unloaded entirely then re-loaded into the system. For this we should:

1. Call Finalizers/Shutdown Function Pointers on the Modules
2. Remove the Module from the Host Process `_modules` Dictionary
3. Do a Memory `Clear()` to remove all segments/variables defined within a Memory Core
4. Call the Host Process `AddModule()` to re-add/initialize the module again.

In theory this would be a clean start for the module and should prevent any memory leaks/corruption form a re-initialization as it's basically a repeat of the startup process.

tuday2:
@enusbaum Closer -- want to get some feedback and help in figuring out why we aren't "restarting" -- modules are reloading correctly now

tuday2:
So it's bombing on RTick -- if I step through it everything works fine, so guessing timer related -- been trying resetting them, clearing them

tuday2:
```System.Collections.Generic.KeyNotFoundException: The given key '15' was not present in the dictionary.
   at System.Collections.Generic.Dictionary`2.get_Item(TKey key)
   at MBBSEmu.Memory.MemoryCore.GetInstruction(UInt16 segment, UInt16 instructionPointer) in C:\Users\JonathanSider\Documents\GitHub\MBBSEmu_Tuday2\MBBSEmu\Memory\MemoryCore.cs:line 274
   at MBBSEmu.HostProcess.ExecutionUnits.ExecutionUnit.Execute(IntPtr16 entryPoint, UInt16 channelNumber, Boolean simulateCallFar, Boolean bypassState, Queue`1 initialStackValues, UInt16 initialStackPointer) in C:\Users\JonathanSider\Documents\GitHub\MBBSEmu_Tuday2\MBBSEmu\HostProcess\ExecutionUnits\ExecutionUnit.cs:line 106
   at MBBSEmu.Module.MbbsModule.Execute(IntPtr16 entryPoint, UInt16 channelNumber, Boolean simulateCallFar, Boolean bypassSetState, Queue`1 initialStackValues, UInt16 initialStackPointer) in C:\Users\JonathanSider\Documents\GitHub\MBBSEmu_Tuday2\MBBSEmu\Module\MBBSModule.cs:line 259
   at MBBSEmu.HostProcess.MbbsHost.Run(String moduleName, IntPtr16 routine, UInt16 channelNumber, Boolean simulateCallFar, Queue`1 initialStackValues) in C:\Users\JonathanSider\Documents\GitHub\MBBSEmu_Tuday2\MBBSEmu\HostProcess\MbbsHost.cs:line 802
   at MBBSEmu.HostProcess.MbbsHost.WorkerThread() in C:\Users\JonathanSider\Documents\GitHub\MBBSEmu_Tuday2\MBBSEmu\HostProcess\MbbsHost.cs:line 317
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()```

**Comment by tuday2** [comment]
@enusbaum How would I get APIREPORT to exit when its done?  I had to do it a little differently as it can do multiple modules at once

the other code is cleaned up -- probably would be good to get a review

File: MBBSEmu/HostProcess/MbbsHost.cs [hunk 2]
```diff
@@ -91,13 +93,18 @@ public class MbbsHost : IMbbsHost
         /// </summary>
         private readonly Timer _timer;
 
+        // Timer to prevent main loop from executing while cleanup is running
+        private readonly Timer _nightlyCleanupTimer;
+
+        private bool _performingNightlyCleanup;
```
enusbaum, line 99: Add a comment/XML comment on the context for this variable as it's similar to others that are declared

MBBSEmu/Program.cs
```diff
@@ -262,24 +264,17 @@ static void Main(string[] args)
                     return;
                 }
 
+                //Setup and Run Host
+                var host = _serviceResolver.GetService<IMbbsHost>();
+                host.Start(_moduleConfigurations);
```
enusbaum, line 269: Move API Reporting before calling Start() and you won't have to worry about calling Stop() within the report

MBBSEmu/Module/ModuleConfiguration.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// <summary>
+    ///     Module Configuration
+    /// </summary>
+        
+    public class ModuleConfiguration
+    {
+        /// <summary>
+        ///     Module "Identifier" from moduleConfig.json or commandline
+        /// </summary>
+        public string ModIdentifier { get; set; }
```
enusbaum, line 12: I would rather these be `ModuleIdentifier` or just `Identifier`, as the implied context is it's already in relation to a module.

enusbaum (changes requested):
Minor changes but looks good!

MBBSEmu/HostProcess/MbbsHost.cs
tuday2: Complete.

MBBSEmu/Program.cs
tuday2: The problem is that the modules aren't initialized until we get into host now -- so we have to host.Start(_moduleConfiguration) or else the APIREPORT won't be able to run (no modules)

I think APIREPORT depends on knowing where all the files are etc. which requires MBBSModule to be constructed -- unless we want to look at changing it?

MBBSEmu/Module/ModuleConfiguration.cs
tuday2: Changed. Love "Refactor->Rename" :)

MBBSEmu/Program.cs
tuday2: Moved Stop() to program.cs per your suggestion.

tuday2 (commented):
Changes complete.

MBBSEmu.Tests/Integration/MBBSEmuIntegrationTestBase.cs
```diff
@@ -73,10 +73,12 @@ protected void ExecuteTest(TestLogic testLogic)
```
paladine, line 73: `var modules` is no longer used right?

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -91,13 +93,24 @@ public class MbbsHost : IMbbsHost
...
+        
         /// <summary>
         ///     Flag that controls whether the main loop will perform a nightly cleanup
         /// </summary>
         private bool _performCleanup = false;
 
         private readonly IGlobalCache _globalCache;
         private readonly IFileUtility _fileUtility;
+        private List<ModuleConfiguration> _moduleConfigurations;
```
paladine, line 113: `readonly` too if possible

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -119,18 +132,39 @@ public MbbsHost(ILogger logger, IGlobalCache globalCache, IFileUtility fileUtili
...
+        public void Start(List<ModuleConfiguration> moduleConfigurations)
         {
+            //Load Modules
+            foreach (var m in moduleConfigurations)
+                AddModule(new MbbsModule(_fileUtility, _logger, m.ModuleIdentifier, m.ModulePath) {MenuOptionKey = m.MenuOptionKey});
+
+            //Remove any modules that did not properly initialize
+            foreach (var (_, value) in _modules.Where(m => m.Value.EntryPoints.Count == 1))
+            {
+                _logger.Warn($"{value.ModuleIdentifier} not propery initialized, Removing");
```
paladine, line 151: properly

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -119,18 +132,39 @@ public MbbsHost(ILogger logger, IGlobalCache globalCache, IFileUtility fileUtili
             _incomingSessions = new Queue<SessionBase>();
             _cleanupTime = ParseCleanupTime();
             _timer = new Timer(unused => _performCleanup = true, this, NowUntil(_cleanupTime), TimeSpan.FromDays(1));
-
+            _nightlyCleanupTimer = new Timer(ProcessNightlyCleanup, this, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
```
paladine, line 135: Is this really needed? I would think running the cleanup routine in the original WorkerThread would be better since it eliminates many of the multithreading issues?

Or is the point to let WorkerThread die so you can create a new one? Though I'm not sure how that's possible since WorkerThread is guarded against `!= null` but is never set back to `null`

MBBSEmu/Program.cs
```diff
@@ -262,24 +264,20 @@ static void Main(string[] args)
...
                 if (_doApiReport)
                 {
-                    foreach (var m in modules)
-                    {
-                        var apiReport = new ApiReport(_logger, m);
-                        apiReport.GenerateReport();
-                    }
+                    host.GenerateAPIReport();
+                    
+                    //APIREPORT done -- Stop Host
```
paladine, line 276: nit: comments that just repeat what the code does are noise and aren't necessary

paladine (changes requested):

MBBSEmu/HostProcess/MbbsHost.cs
tuday2: If I add readonly it won't compile because of the below code:

```
           //Save module configurations for cleanup
            _moduleConfigurations = moduleConfigurations;
```

MBBSEmu/HostProcess/MbbsHost.cs
tuday2: Fixed.

MBBSEmu/HostProcess/MbbsHost.cs
tuday2: This was added to "spin" the worker routine to let the cleanup run without RTICK etc. running , in testing this was the solve to prevent anything from going on while cleanup is running and to not run it again if currently running

```
            _nightlyCleanupTimer = new Timer(ProcessNightlyCleanup, this, TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(1));
```

MBBSEmu/Program.cs
tuday2: Removed

MBBSEmu.Tests/Integration/MBBSEmuIntegrationTestBase.cs
tuday2: Correct, removed from test

tuday2:
@paladine Resolved/Responded to review

MBBSEmu.Tests/Integration/MBBSEmuIntegrationTestBase.cs
```diff
@@ -66,11 +66,9 @@ protected void ExecuteTest(TestLogic testLogic)
 
             CopyModuleToTempPath(ResourceManager.GetTestResourceManager());
 
-            var modules = new List<MbbsModule>
-            {
-                new MbbsModule(serviceResolver.GetService<IFileUtility>(), serviceResolver.GetService<ILogger>(), "MBBSEMU", _modulePath)
-            };
-
+            new MbbsModule(serviceResolver.GetService<IFileUtility>(), serviceResolver.GetService<ILogger>(), "MBBSEMU",
+                _modulePath);
```
paladine, line 70: still not needed though?

paladine (approved):

MBBSEmu.Tests/Integration/MBBSEmuIntegrationTestBase.cs
tuday2: Here is what Host does:
```
 //Load Modules
            foreach (var m in moduleConfigurations)
                AddModule(new MbbsModule(_fileUtility, _logger, m.ModuleIdentifier, m.ModulePath) {MenuOptionKey = m.MenuOptionKey});
```

Basically MBBSModule initialize the module (loads msgs/mcvs/etc.) and AddModule activates it so to speak.... so yes we do need to call MBBSModule, and then in Host we call "AddModule" to initialize it

PR #147, opened 2020-09-25
Tests for Borland math functions & various bug fixes

paladine:
Change OUTBSZ to 8192 while I was at it


MBBSEmu/HostProcess/ExportedModules/Galgsbl.cs
```diff
@@ -24,6 +24,8 @@ public class Galgsbl : ExportedModuleBase, IExportedModule, IStoppable
         /// <returns></returns>
         public const ushort Segment = 0xFFFE;
 
+        public const ushort OUTBUF_SIZE = 8192;
```
enusbaum, line 27: We can move this constant to the base class so the ExportedModules don't reference one another

enusbaum (changes requested):
Just the one change to break the dependency on GSBL<->MAJORBBS

PR #163, opened 2020-10-05
Retrieve the physical record length from the btrieve .DAT file

paladine:
Instead of trying to synthesize it.

Will likely invalidate most pre-existing .EMU files.

### DIFF
--- a/MBBSEmu/Btrieve/BtrieveFile.cs
+++ b/MBBSEmu/Btrieve/BtrieveFile.cs
@@ -42,7 +42,6 @@ namespace MBBSEmu.Btrieve
             }
         }
 
-
         private ushort _recordLength;
         /// <summary>
         ///     Defined Length of the records within the Btrieve File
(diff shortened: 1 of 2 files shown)


MBBSEmu/Btrieve/BtrieveFile.cs
```diff
@@ -65,6 +64,28 @@ public ushort RecordLength
             }
         }
 
+        private ushort _physicalRecordLength;
+        /// <summary>
+        ///     Actual Length of the records within the Btrieve File, including additional padding.
+        /// </summary>
+        public ushort PhysicalRecordLength
```
enusbaum, line 71: Should we [JsonIgnore] this since it's only needed on initial load?

enusbaum (approved):
Minor suggestion -- merge if you think it's not needed. Glad you were able to solve this mystery! 👍

MBBSEmu/Btrieve/BtrieveFile.cs
paladine: I'm inclined to keep it since it serves as useful debugging information when you look at the `.EMU` file

PR #185, opened 2020-10-10
Update Global Formatting / Prompt

tuday2:
*Added padding to align user list for /# global
*Corrected padding on menu to account for # of modules and actually format correctly
*Redisplay prompt after global is executed to behave more like MBBS

### DIFF
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -217,10 +217,13 @@ namespace MBBSEmu.HostProcess
                         {
                             session.Status = 1;
                             session.InputBuffer.SetLength(0);
+
+                            //Redisplay Main Menu prompt after global if session is at Main Menu
+                            if (session.SessionState == EnumSessionState.MainMenuInput)
+                            { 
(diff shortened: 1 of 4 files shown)


MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -217,6 +217,10 @@ private void WorkerThread()
                         {
                             session.Status = 1;
                             session.InputBuffer.SetLength(0);
+                            
+                            //Redisplay Main Menu prompt after global
```
enusbaum, line 221: We should split this display out from `MainMenuDisplay()` into a new method with a new state such as `MainMenuInputDisplay()` that displays this string. This way we can set the user state here to display it and we're not duplicating the string.

Additionally it'd let us perhaps add the ability for someone to implement a custom version using a specified ANSI in a future enhancement.

enusbaum (changes requested):
Recommending a refactor to reduce duplicating the string

MBBSEmu/HostProcess/MbbsHost.cs
tuday2: Great idea... this is done.

enusbaum (approved):

PR #186, opened 2020-10-12
Move Shutdown logic into worker thread to minimize multithread issues

paladine:
Includes a unit test for the shutdown routine


enusbaum (approved):
LGTM -- also I think there's good future potential for potentially removing some of the loop logic and handling STT events using the new event handler. Food for thought!

Issue #181, opened 2020-10-10
[SFABLX] BLOX - Module crashes when resetting high scores


tuday2:
**Module Information**
* Identifier: SFABLX
* Module Name: BLOX
* Version: 1.01

**Describe the bug**
Module crashes when resetting high scores

**To Reproduce**
1. Load Module
2. Enter Module
3. Select "R"eset High Scores

**Expected behavior**
High scores are reset

**Screenshots**
```
2020-10-10 07:25:41.0053 Error MBBSEmu.HostProcess.ExportedModules.Majorbbs.Invoke Unknown Exported Function Ordinal in MAJORBBS: 211:FILELENGTH
Unhandled exception. System.ArgumentOutOfRangeException: Specified argument was out of the range of valid values. (Parameter 'Unknown Exported Function Ordinal in MAJORBBS: 211:FILELENGTH')
   at MBBSEmu.HostProcess.ExportedModules.Majorbbs.Invoke(UInt16 ordinal, Boolean offsetsOnly)
   at MBBSEmu.HostProcess.ExecutionUnits.ExecutionUnit.ExternalFunctionDelegate(UInt16 ordinal, UInt16 functionOrdinal)
   at MBBSEmu.HostProcess.ExecutionUnits.ExecutionUnit.Execute(IntPtr16 entryPoint, UInt16 channelNumber, Boolean simulateCallFar, Boolean bypassState, Queue`1 initialStackValues, UInt16 initialStackPointer)
   at MBBSEmu.Module.MbbsModule.Execute(IntPtr16 entryPoint, UInt16 channelNumber, Boolean simulateCallFar, Boolean bypassSetState, Queue`1 initialStackValues, UInt16 initialStackPointer)
   at MBBSEmu.HostProcess.MbbsHost.WorkerThread()
   at System.Threading.ThreadHelper.ThreadStart_Context(Object state)
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
--- End of stack trace from previous location where exception was thrown ---
   at System.Threading.ExecutionContext.RunInternal(ExecutionContext executionContext, ContextCallback callback, Object state)
   at System.Threading.ThreadHelper.ThreadStart()
```

**Software Information:**
- OS: Windows Server 2019 Standard
- Telnet Client: NetRunner
- MBBSEmu Build: 100920-8

enusbaum:
[...]

PR #187, opened 2020-10-12
Added Ordinal 211 filelength & test for it

tuday2:


tuday2:
Closes #181

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -7699,5 +7703,28 @@ private void stp4cs()
...
+        ///     filelength - gets file size in bytes
+        ///
+        ///     Signature: long filelength(int handle);
+        /// </summary>
+        private void filelength()
+        {
+            var fileHandle = GetParameter(0);
+            
+            if (FilePointerDictionary.TryGetValue(fileHandle,out var filePointer))
+            {
```
enusbaum, line 7717: Before the `if`, you can declare `var result = -1L` and if the value exists, set `result = filePointer.Length` (no need to cast as `result` will already be a `long`).

This way you can just have a single statement setting the registers to the value of `result`. Reduces complexity of the method 👍

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -14,6 +14,7 @@
 using System;
 using System.Collections.Generic;
 using System.ComponentModel;
+using System.ComponentModel.Design;
```
enusbaum, line 17: Can probably remove this

enusbaum (changes requested):
Great Test! Minor changes on the Ordinal

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
tuday2: Nice ok, that made is a lot simpler 👍

enusbaum (approved):

PR #191, opened 2020-10-13
Update MenuRoutines.cs fix sorting >9 modules

tuday2:
One last tweak to fix numbers being sorted as strings -- padding 0's to get numbers in correct order.

Modules with custom MenuOptionKeys are sorted as their ascii values but figured if someone is using that they can customize all of them

Screenshot 2020-10-13 001708


tuday2:
**After change:**

Screenshot 2020-10-13 002529

enusbaum (approved):

PR #198, opened 2020-10-16
Memory performance optimizations

enusbaum:
Memory performance optimizations by reducing Framework calls on simple operations.

Optimizations took benchmark locally on an i9-9900k from 9 MIP to 12 MIPS in Release Mode.

### DIFF
--- a/MBBSEmu/Memory/IntPtr16.cs
+++ b/MBBSEmu/Memory/IntPtr16.cs
@@ -10,13 +10,21 @@ namespace MBBSEmu.Memory
         public ushort Segment
         {
             get => BitConverter.ToUInt16(Data, 2);
-            set => Array.Copy(BitConverter.GetBytes(value), 0, Data, 2, 2);
+            set
+            {
+                Data[2] = (byte)(value & 0xFF);
+                Data[3] = (byte)(value >> 8);
+            }
 
         }
         public ushort Offset
         {
             get => BitConverter.ToUInt16(Data, 0);
-            set => Array.Copy(BitConverter.GetBytes(value), 0, Data, 0, 2);
+            set
+            {
+                Data[0] = (byte)(value & 0xFF);
+                Data[1] = (byte)(value >> 8);
+            }
         }
 
         public readonly byte[] Data = new byte[Size];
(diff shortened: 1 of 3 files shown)


MBBSEmu/Memory/IntPtr16.cs
```diff
@@ -10,13 +10,21 @@ public class IntPtr16 : IEquatable<IntPtr16>
...
             get => BitConverter.ToUInt16(Data, 0);
-            set => Array.Copy(BitConverter.GetBytes(value), 0, Data, 0, 2);
+            set
+            {
+                Data[0] = (byte)(value & 0xFF);
+                Data[1] = (byte)(value >> 8);
+            }
         }
 
         public readonly byte[] Data = new byte[Size];
```
paladine, line 30: I wonder if this would be even more efficient if it were an array of ushort, so you could read/write a word at a time?

MBBSEmu/Memory/MemoryCore.cs
```diff
@@ -476,8 +476,13 @@ public void SetByte(ushort segment, ushort offset, byte value)
         /// <param name="segment"></param>
         /// <param name="offset"></param>
         /// <param name="value"></param>
-        public void SetWord(ushort segment, ushort offset, ushort value) =>
-            SetArray(segment, offset, BitConverter.GetBytes(value));
+        public void SetWord(ushort segment, ushort offset, ushort value)
+        {
+            _ = new Span<byte>(_memorySegments[segment])
```
paladine, line 481: does this allocate a new object or is this something like a stack variable?

MBBSEmu/Memory/IntPtr16.cs
enusbaum: I'll benchmark changing the getter and setter for Data so the segment/offset use a backing ushort and Data is built on request from those values.

I suspect we use .Segment and .Offset way more frequently than .Data, which I believe is only used when we're committing the value to memory.

MBBSEmu/Memory/MemoryCore.cs
enusbaum: Span<T> lives on the stack and gives us a ref to the dictionary value. This avoids multiple calls to the dictionary indexer which performs a hash lookup.

PR #208, opened 2020-10-18
Optimizations & Fixing DIV

enusbaum:
- Moved information only used for debugging into preprocessor
- Removed unnecessary local variables
- Fixed DIV/IDIV and added Unit Tests

Benchmark on this branch gets 13.8-14.2 MIPS -- depending on which core it gets assigned to on my development machine (i9-9900k)

### DIFF
--- a/MBBSEmu/CPU/CPUCore.cs
+++ b/MBBSEmu/CPU/CPUCore.cs
@@ -231,17 +231,18 @@ namespace MBBSEmu.CPU
         public void Tick()
         {
             //Check for segment end
-            if ((Registers.CS == ushort.MaxValue || Registers.CS == 0) && (Registers.IP == ushort.MaxValue || Registers.IP == 0))
+            if (Registers.CS == ushort.MaxValue)
             {
                 Registers.Halt = true;
                 return;
             }
 
+#if DEBUG
             _currentInstructionPointer.Offset = Registers.IP;
             _currentInstructionPointer.Segment = Registers.CS;
+#endif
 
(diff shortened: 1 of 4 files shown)


MBBSEmu/CPU/CPUCore.cs
```diff
@@ -231,17 +231,18 @@ public void Push(ushort value)
         public void Tick()
         {
             //Check for segment end
-            if ((Registers.CS == ushort.MaxValue || Registers.CS == 0) && (Registers.IP == ushort.MaxValue || Registers.IP == 0))
+            if (Registers.CS == ushort.MaxValue)
```
paladine, line 234: couldn't you add a sentinel instruction at the end of memory which when executed kills the core? Just throwing out ideas

paladine (approved):

MBBSEmu/CPU/CPUCore.cs
enusbaum: Technically the program can `retf` from any point. Methods that make use of [longjmp](https://en.cppreference.com/w/cpp/utility/program/longjmp) are an example of this, where it can call a final `retf` from a method that it didn't start in.

To solve for this, we push `0xFFFF` to the stack before each method call, so on the "final" `reft`, a CS value of `0xFFFF` is popped from the stack and we know we're at the end.

https://github.com/enusbaum/MBBSEmu/blob/master/MBBSEmu/HostProcess/ExecutionUnits/ExecutionUnit.cs#L86

The simplification of this if statement is because the other cases (checking for 0) were symptoms of stack bugs previously. It should always finish on CS `0xFFFF`

Issue #209, opened 2020-10-19
Improve error messages with invalid appsettings.json values


paladine:
Generally the system is intolerant of missing/invalid appsettings.json values. If you're missing `Database.File`, you get some random message about missing the database but no guidance on how to resolve it. 

If missing `GSBL.Activation`, it just crashes with no guidance at all.

Audit all the settings and make sure that the program exits gracefully with useful instructions.

tuday2:
Close Me @enusbaum

enusbaum:
Fixed with #215

PR #215, opened 2020-10-24
JSON Appsettings refactor

tuday2:
Issue #209
Submitting for initial review, still have more error messages to add


MBBSEmu/Configuration.cs
```diff
@@ -0,0 +1,136 @@
+﻿using System;
+using System.IO;
+using Microsoft.Extensions.Configuration;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+
+namespace MBBSEmu
+{
+    public static class AppSettings
```
enusbaum, line 9: Let's not make this static, make it an instanced class that we can pass into classes using Dependency Injection

MBBSEmu/Configuration.cs
```diff
@@ -0,0 +1,136 @@
...
+        public static bool RloginPortPerModule => GetBoolAppSettings("Rlogin.PortPerModule");
+        public static string DatabaseFile => GetStringAppSettings("Database.File");
+        
+        //Optional Keys
+        public static string ANSILogin => ConfigurationRoot["ANSI.Logon"];
+        public static string ANSILogoff => ConfigurationRoot["ANSI.Logoff"];
+        public static string ANSISignup => ConfigurationRoot["ANSI.Signup"];
+        public static string ANSIMenu => ConfigurationRoot["ANSI.Menu"];
+
+        private static string GetStringAppSettings(string key)
```
enusbaum, line 49: We can probably combine these methods using a `GetAppSettings<T>(string key)`

enusbaum (changes requested):

MBBSEmu/Configuration.cs
tuday2: Fixed.

tuday2:
Ready for review

enusbaum (approved):

PR #222, opened 2020-10-28
Flag Optimizations

enusbaum:
- Small Optimization/Refactor for CPU Flags
- Removed them from a dedicated class & moved flag evaluation routines into CPU Core

Nets about a 5% speed bump in the benchmark on my local machine.


MBBSEmu/CPU/CPUCore.cs
```diff
@@ -2051,8 +2051,8 @@ private void Op_Test_8()
             var source = GetOperandValueUInt8(_currentInstruction.Op1Kind, EnumOperandType.Source);
 
             destination &= source;
-            Registers.F.Evaluate(EnumFlags.ZF, destination);
-            Registers.F.Evaluate(EnumFlags.SF, destination);
+            Flags_EvaluateZero(destination);
```
paladine, line 2054: these always seem to come in pairs, perhaps combine the zero/sign evaluation into one method call for some minor gains?

MBBSEmu/CPU/CPUCore.cs
```diff
@@ -2080,9 +2080,6 @@ private void Op_Cmp()
                 case 2:
                     Op_Sub_16();
                     return;
-                case 4:
```
paladine, line 2083: did we lose this case?

paladine (approved):

MBBSEmu/CPU/CPUCore.cs
enusbaum: I reviewed every opcode which modifies both the sign or the zero flag and they're always in tandem. I checked in the modified method which evaluates the value and sets both flags accordingly.

MBBSEmu/CPU/CPUCore.cs
enusbaum: Oh -- I stripped out the 32-bit branches I had started to write. Those snuck in because I was accidentally testing a MajorMUD version for Worldgroup 3.0 DOS.

MBBSEmu/CPU/CPUCore.cs
```diff
@@ -3536,43 +3509,24 @@ private void Op_Fdivrp()
...
         {
-            if (result == 0)
-            {
-                Registers.F = Registers.F.SetFlag((ushort)EnumFlags.ZF);
-            }
-            else
-            {
-                Registers.F = Registers.F.ClearFlag((ushort)EnumFlags.ZF);
-            }
+            Registers.F = result.IsNegative() ? Registers.F.SetFlag((ushort)EnumFlags.SF) : Registers.F.ClearFlag((ushort)EnumFlags.SF);
```
paladine, line 3518: if result.IsNegative() you can skip evaluation of Zero since it's obviously not 0

e.g.

```
if (result.IsNegative()) {
  setFlag(SF);
  clearFlag(ZF);
} else {
  clearFlag(SF);
  if (result == 0) .... set | clear
}
```

paladine (approved):

PR #235, opened 2020-10-30
Set sane defaults for new MemoryStreams to avoid reallocs

paladine:
Plus small tweak to fix up those ``\r\n`` in McvFile :)

### DIFF
--- a/MBBSEmu/Module/McvFile.cs
+++ b/MBBSEmu/Module/McvFile.cs
@@ -15,7 +15,7 @@ namespace MBBSEmu.Module
     /// </summary>
     public class McvFile
     {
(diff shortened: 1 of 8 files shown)


MBBSEmu/Module/McvFile.cs
```diff
@@ -15,6 +15,8 @@ namespace MBBSEmu.Module
     /// </summary>
     public class McvFile
     {
+        private static readonly byte[] NEW_LINE = new byte[] {(byte)'\r', (byte)'\n'};
```
enusbaum: You can use an array initializer here, no need for the `new byte[]`

enusbaum (approved):

PR #258, opened 2020-11-05
Add heartbeat and # channels config

tuday2:
--Need to add a few more config defaults

### DIFF
--- a/MBBSEmu/Session/Telnet/TelnetSession.cs
+++ b/MBBSEmu/Session/Telnet/TelnetSession.cs
@@ -14,7 +14,7 @@ namespace MBBSEmu.Session.Telnet
     {
         private int _iacPhase;
 
-        private static byte[] IAC_NOP = { 0xFF, 0xF1};
+        private static readonly byte[] IAC_NOP = { 0xFF, 0xF1};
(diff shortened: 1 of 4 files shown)


MBBSEmu/Session/Telnet/TelnetSession.cs
```diff
@@ -41,13 +41,20 @@ private class TelnetOptionsValue {
...
         {
             SessionType = EnumSessionType.Telnet;
             SessionState = EnumSessionState.Unauthenticated;
 
             _iacFilter = new IacFilter(logger);
             _iacFilter.IacVerbReceived += OnIacVerbReceived;
+
+            //Set Telnet Heartbeat from Config
+            _configuration = configuration;
+            if (!_configuration.TelnetHeartbeat)
```
enusbaum, line 54: I would modify the overridden `Heartbeat` method to put the `if` statement there to not call the `base.Send` if this is set to false.

https://github.com/enusbaum/MBBSEmu/blob/e518a8b0251eea93c316d5f13f178e19dabb6ffe/MBBSEmu/Session/Telnet/TelnetSession.cs#L69

enusbaum (changes requested):

enusbaum (approved):

PR #269, opened 2020-11-07
Add `/SYSOP` Global Commands

enusbaum:
- Added Key Management to SYSOP Commands

Allows users with the `SYSOP` key to manage keys for other users.


MBBSEmu/HostProcess/GlobalRoutines/SysopGlobal.cs
```diff
@@ -0,0 +1,202 @@
+﻿using MBBSEmu.Database.Repositories.Account;
```
paladine, line 1: approving now, but would be swell to have some unit tests around these globals

paladine (approved):

PR #275, opened 2020-11-08
Update strcmp to check empty

tuday2:
Fixes Mutants! -- UACOFF needs to be further explored


paladine (changes requested):
plz add relevant test in stricmp_Tests.cs

paladine (changes requested):
str**i**cmp_Tests :)

paladine (approved):

PR #283, opened 2020-11-11
Refactor Btrieve ordinals

paladine:
Created a new method `PerformOperation` which basically handles any `EnumBtrieveOperationCode`

`UpdateBB` method in `Majorbbs.cs` handles the updating of the BB struct and eliminates all the code duplication.

### DIFF
--- a/MBBSEmu/Btrieve/Enums/EnumBtrieveOperationCodes.cs
+++ b/MBBSEmu/Btrieve/Enums/EnumBtrieveOperationCodes.cs
@@ -122,7 +122,7 @@
         public static bool UsesPreviousQuery(this EnumBtrieveOperationCodes code)
         {
             var memberInstance = code.GetType().GetMember(code.ToString());
-            if (memberInstance == null || memberInstance.Length <= 0) return false;
+            if (memberInstance.Length <= 0) return false;
 
             return System.Attribute.GetCustomAttribute(memberInstance[0], typeof(UsesPreviousQuery)) != null;
         }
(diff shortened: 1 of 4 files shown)


MBBSEmu/Btrieve/Enums/EnumBtrieveOperationCodes.cs
```diff
@@ -1,48 +1,138 @@
...
 
         None = ushort.MaxValue
     }
+
+    public static class Extensions
+    {
+        public static bool UsesPreviousQuery(this EnumBtrieveOperationCodes code)
+        {
+            var memberInstance = code.GetType().GetMember(code.ToString());
+            if (memberInstance == null || memberInstance.Length <= 0) return false;
```
enusbaum, line 125: `memberInstance` will never be null, just an empty array so we can remove this check

MBBSEmu/Btrieve/Enums/EnumBtrieveOperationCodes.cs
```diff
@@ -1,48 +1,138 @@
...
+
+            return System.Attribute.GetCustomAttribute(memberInstance[0], typeof(UsesPreviousQuery)) != null;
+        }
+
+        public static bool AcquiresData(this EnumBtrieveOperationCodes code)
+        {
+            var memberInstance = code.GetType().GetMember(code.ToString());
+            if (memberInstance == null || memberInstance.Length <= 0) return false;
```
enusbaum, line 133: same as above

enusbaum (approved):
Boss refactor!

Only the two comparisons that will never evaluate true.

PR #289, opened 2020-11-13
Issue 286

### LINKED ISSUE #286: [HVSXROAD] Crossroads of the Elements - High score board won't populate after cleanup
**Module Information**
* Identifier [HVSXROAD]
* Module Name [Crossroads of the Elements]
* Version [1.00j]

**Describe the bug**
I've added a few screenshots to show the issue.  The high score board stays blank, even after players reach above level 1/1.

**To Reproduce**
1. Start Crossroads (as a NORMAL user)
2. Level up
3. Wait until cleanup
4. Observe the score board

**Expected behavior**
The score board is populated with non-sysop users that are above level 1/1.

**Screenshots**
**You can see the nightly cleanup event and reference the work being done with HVSXROAD**
Screenshot from 2020-11-12 06-32-15

**This is what the high score board looks like if no users in the system qualify to be on the board. E.g., not greater than level 1/1, or a sysop.**
Screenshot from 2020-11-12 06-31-42

**Here is a player that is level 1/30. You can see the scoreboard after the cleanup changed.  The "<None!>" descriptors are gone, indicating the system knows a user qualifies, but simply isn't being populated.**
Screenshot from 2020-11-12 06-30-08

**Software Information:**
- OS: [Fedora 32]
- Telnet Client [Linux telnet and Telix via DOSBox]

enusbaum:
- Added Implementation of `uidkey()`
- Fixed `f_printf()` to not write the null terminator of the string being written to the file stream
- Fixed `f_printf()` to return the number of bytes written
- Unit Tests for Both

Fixes #286

### DIFF
--- a/MBBSEmu.Tests/ExportedModules/Majorbbs/FileTestBase.cs
+++ b/MBBSEmu.Tests/ExportedModules/Majorbbs/FileTestBase.cs
@@ -76,6 +76,54 @@ namespace MBBSEmu.Tests.ExportedModules.Majorbbs {
             return mbbsEmuCpuRegisters.AX;
         }
 
+        protected ushort f_printf(IntPtr16 filep, string formatString, params object[] values)
+        {
+            var fprintfParameters = new List<ushort> {filep.Offset, filep.Segment};
+
+            //Add Formatted String
+            var inputStingParameterPointer = mbbsEmuMemoryCore.AllocateVariable(Guid.NewGuid().ToString(), (ushort)(formatString.Length + 1));
+            mbbsEmuMemoryCore.SetArray(inputStingParameterPointer, Encoding.ASCII.GetBytes(formatString));
+            fprintfParameters.Add(inputStingParameterPointer.Offset);
+            fprintfParameters.Add(inputStingParameterPointer.Segment);
+
+            //Add Parameters
+            foreach (var v in values)
+            {
+                switch (v)
+                {
+                    case string @parameterString:
+                    {
+                        var stringParameterPointer = mbbsEmuMemoryCore.AllocateVariable(Guid.NewGuid().ToString(), (ushort)(@parameterString.Length + 1));
+                        mbbsEmuMemoryCore.SetArray(stringParameterPointer, Encoding.ASCII.GetBytes(@parameterString));
+                        fprintfParameters.Add(stringParameterPointer.Offset);
+                        fprintfParameters.Add(stringParameterPointer.Segment);
+                        break;
+                    }
+                    case uint @parameterULong:
+                    {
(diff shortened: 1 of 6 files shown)


MBBSEmu.Tests/ExportedModules/Majorbbs/FileTestBase.cs
```diff
@@ -76,6 +76,54 @@ protected ushort fwrite(IntPtr16 destPtr, ushort size, ushort count, IntPtr16 fi
...
+                        break;
+                    }
+                    case int @parameterLong:
+                    {
+                        var longBytes = BitConverter.GetBytes(@parameterLong);
+                        fprintfParameters.Add(BitConverter.ToUInt16(longBytes, 0));
+                        fprintfParameters.Add(BitConverter.ToUInt16(longBytes, 2));
+                        break;
+                    }
+                    case ushort @parameterInt:
```
paladine, line 116: `@parameterUInt`

Also, add the `short` case just to cover all the bases?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -3882,6 +3882,8 @@ private void f_printf()
 #if DEBUG
             _logger.Info($"Wrote {formattedMessage.Length} bytes to {fileStructPointer} (Stream: {fileStruct.curp})");
 #endif
+
+            Registers.AX = (ushort)formattedMessage.Length;
```
paladine, line 3886: nice catch!

MBBSEmu.Tests/ExportedModules/Majorbbs/key_Tests.cs
```diff
@@ -99,6 +100,29 @@ public void haskey_Empty_Test()
 
         }
 
+        [Fact]
+        public void uidkey_Test()
```
paladine, line 104: let's have other tests for
-  empty `accountKey`
- non-default key (i.e. insert a new key for the user, query specifically on it)
- unknown user with default key = true
- unknown user with custom key = false

paladine (changes requested):

MBBSEmu.Tests/ExportedModules/Majorbbs/FileTestBase.cs
enusbaum: I pulled this from the `prf` tests. There might be a way to refactor this into the base as to not have multiple copies of the logic. Let me see what we can do there.

enusbaum: To MBBSEMu, they're longs ;)

paladine (approved):

PR #304, opened 2020-11-19
Add random tests

tuday2:


MBBSEmu.Tests/ExportedModules/Majorbbs/random_Tests.cs
```diff
@@ -0,0 +1,84 @@
...
+            //Verify Results
+            Assert.Equal(mbbsEmuCpuRegisters.AX, expectedValue);
+        }
+
+        [Theory]
+        [InlineData(1, 65000)]
+        [InlineData(10000, 65000)]
+        [InlineData(10, 20)]
+        [InlineData(11, 12)]
+        [InlineData(32767, 65500)]
```
paladine, line 47: How about adding some values that are actually long? All these test values have high words = 0

MBBSEmu.Tests/ExportedModules/Majorbbs/random_Tests.cs
```diff
@@ -0,0 +1,84 @@
...
+        [Fact]
+        public void randTest()
+        {
+            Reset();
+
+            ExecuteApiTest(HostProcess.ExportedModules.Majorbbs.Segment,RAND_ORDINAL, new List<ushort>());
+
+            //Verify Results
+            Assert.InRange(mbbsEmuCpuRegisters.AX, 1, 32767);
+        }
```
paladine, line 82: There's no real way to test this exactly, since due to the random nature of the test and the fact that we're not controlling the randomness, this test might pass every single time until finally it fails based on whatever random number is returned.

The only way to control this would be to have the test control the `Random` used by the `Majorbbs.cs`. So you'd have to have some sort of test-only method where the test passes the `Random` to use, and then controls the values spit out by the `Random`

paladine: This is outside the scope of this review, but a good candidate for a follow-on

paladine (changes requested):

paladine (approved):

PR #308, opened 2020-11-22
Create itoa, memset, movmem Tests -- Update desc movmem Majorbbs.cs

tuday2:


tuday2:
Updated Description of movmem and renamed (was movemem) -- it was referencing ordinal 410 which is not implemented yet (memmove)

MBBSEmu.Tests/ExportedModules/Majorbbs/itoa_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+                new List<ushort>
+                {
+                    intConvert,
+                    destinationStringPointer.Offset,
+                    destinationStringPointer.Segment,
+                    baseInt
+                });
+
+            //Verify Results
+            Assert.Equal(expected, Encoding.ASCII.GetString(mbbsEmuMemoryCore.GetString("STRDST", true)));
```
paladine, line 39: we'll want to assert on `DX:AX` as well

MBBSEmu.Tests/ExportedModules/Majorbbs/memset_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        private const int MEMSET_ORDINAL = 411;
+
+        [Theory]
+        [InlineData(new byte[] { 0x4, 0x4 }, 0x3, 2, new byte[] { 0x3, 0x3 })]
+        [InlineData(new byte[] { 0x4, 0x4, 0x3, 0x2 }, 0x3, 4, new byte[] { 0x3, 0x3, 0x3, 0x3 })]
+        [InlineData(new byte[] { 0x4, 0x4, 0x3, 0x2, 0x1, 0x0 }, 0xE, 6, new byte[] { 0xE, 0xE, 0xE, 0xE, 0xE, 0xE })]
+        [InlineData(new byte[] { 0x4 }, 0x0, 1, new byte[] { 0x0 })]
+        [InlineData(new byte[] { 0xE, 0xF, 0x2, 0x8 }, 0xA, 4, new byte[] { 0xA, 0xA, 0xA, 0xA })]
+        public void memset_Test(byte[] bufMem, ushort valueToFill, ushort memSize, byte[] expected)
```
paladine, line 16: `byte[] expected` isn't necessary since it's always valueToFill * memSize

MBBSEmu.Tests/ExportedModules/Majorbbs/memset_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+    public class memset_Tests : ExportedModuleTestBase
+    {
+        private const int MEMSET_ORDINAL = 411;
+
+        [Theory]
+        [InlineData(new byte[] { 0x4, 0x4 }, 0x3, 2, new byte[] { 0x3, 0x3 })]
+        [InlineData(new byte[] { 0x4, 0x4, 0x3, 0x2 }, 0x3, 4, new byte[] { 0x3, 0x3, 0x3, 0x3 })]
+        [InlineData(new byte[] { 0x4, 0x4, 0x3, 0x2, 0x1, 0x0 }, 0xE, 6, new byte[] { 0xE, 0xE, 0xE, 0xE, 0xE, 0xE })]
+        [InlineData(new byte[] { 0x4 }, 0x0, 1, new byte[] { 0x0 })]
+        [InlineData(new byte[] { 0xE, 0xF, 0x2, 0x8 }, 0xA, 4, new byte[] { 0xA, 0xA, 0xA, 0xA })]
```
paladine, line 15: test with `memSize == 0` would be nice

MBBSEmu.Tests/ExportedModules/Majorbbs/memset_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+                    bufPointer.Offset,
+                    bufPointer.Segment,
+                    valueToFill,
+                    memSize
+                });
+
+            //Verify Results
+            var dstArray = mbbsEmuMemoryCore.GetArray("SETMEMORY", memSize);
+
+            Assert.Equal(expected, dstArray.ToArray());
```
paladine, line 39: and we'll want to assert on `DX:AX` here as well

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -818,7 +818,7 @@ public ReadOnlySpan<byte> Invoke(ushort ordinal, bool offsetsOnly = false)
                     parsin();
                     break;
                 case 420:
```
paladine, line 820: https://rb.gy/z6ham7

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -4058,14 +4058,11 @@ private void parsin()
...
-        ///     Copying takes place as if an intermediate buffer were used, allowing the destination and source to overlap
+        ///     Move a block of memory
         ///
-        ///     Signature: void * memmove ( void * destination, const void * source, size_t num )
-        ///
-        ///     More Info: http://www.cplusplus.com/reference/cstring/memmove/
+        ///     Signature: void movmem(char *source, char *destination, unsigned nbytes)
         /// </summary>
-        private void movemem()
+        private void movmem()
```
paladine, line 4065: technically `memmove()` which we ought to use

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -4058,14 +4058,11 @@ private void parsin()
...
         {
             var sourcePointer = GetParameterPointer(0);
             var destinationPointer = GetParameterPointer(2);
```
paladine, line 4068: we don't set the return value in the ordinal either :(
it returns `dst`

MBBSEmu.Tests/ExportedModules/Majorbbs/movmem_Tests.cs
```diff
@@ -0,0 +1,45 @@
...
+{
+    public class movmem_Tests : ExportedModuleTestBase
+    {
+        private const int MOVMEM_ORDINAL = 420;
+
+        [Theory]
+        [InlineData(3, new byte[] { 0x4, 0x4, 0x2 })]
+        [InlineData(1, new byte[] { 0xE })]
+        [InlineData(2, new byte[] { 0xD, 0xA })]
+        [InlineData(5, new byte[] { 0xD, 0xA, 0xD, 0xD, 0x2 })]
```
paladine, line 15: `moveLength` is always `expected.Length`, is the parameter necessary then?

MBBSEmu.Tests/ExportedModules/Majorbbs/movmem_Tests.cs
```diff
@@ -0,0 +1,45 @@
...
+        public void movmem_Test(ushort moveLength, byte[] expected)
+        {
+            //Reset State
+            Reset();
+
+            //Set Argument Values to be Passed In
+            var srcPointer = mbbsEmuMemoryCore.AllocateVariable("SRC", moveLength);
+            mbbsEmuMemoryCore.SetArray("SRC", expected);
+
+            var dstPointer = mbbsEmuMemoryCore.AllocateVariable("DST", (ushort)(moveLength + 1));
```
paladine, line 26: so the real value of this function is when src & dst are the same buffer. you want to verify that the move works without clobbering all the data, e.g.

char buf[] = "test";
movemem(/* dst= */ buf, /* src= */ buf + 1, 3);
// buf now "estt";

MBBSEmu.Tests/ExportedModules/Majorbbs/movmem_Tests.cs
```diff
@@ -0,0 +1,45 @@
...
+                    srcPointer.Segment,
+                    dstPointer.Offset,
+                    dstPointer.Segment,
+                    moveLength
+                });
+
+            //Verify Results
+            var dstArray = mbbsEmuMemoryCore.GetArray("DST", moveLength);
+
+            Assert.Equal(expected, dstArray.ToArray());
```
paladine, line 42: + assert on `DX:AX`

paladine (changes requested):

MBBSEmu.Tests/ExportedModules/Majorbbs/itoa_Tests.cs
tuday2: DX:AX is 0 because we don't return anything right?

paladine: itoa returns a `char *`, so this should be `Assert.NotEqual(0)`

The ordinal needs to return `dest`

MBBSEmu.Tests/ExportedModules/Majorbbs/memset_Tests.cs
```diff
@@ -34,8 +35,15 @@ public void memset_Test(byte[] bufMem, ushort valueToFill, ushort memSize, byte[
...
 
             //Verify Results
+            var expected = new byte[memSize];
+
+            for (var i = 0; i < memSize; i++)
+                expected[i] = (byte) valueToFill;
+            
             var dstArray = mbbsEmuMemoryCore.GetArray("SETMEMORY", memSize);
 
+            Assert.Equal(0, mbbsEmuCpuRegisters.AX);
```
paladine, line 45: memset should return the first argument as well (`dest`)

MBBSEmu.Tests/ExportedModules/Majorbbs/movmem_Tests.cs
```diff
@@ -33,12 +32,50 @@ public void movmem_Test(ushort moveLength, byte[] expected)
...
                     dstPointer.Offset,
                     dstPointer.Segment,
+                    (ushort) expected.Length
+                });
+
+            //Verify Results
+            var dstArray = mbbsEmuMemoryCore.GetArray("DST", (ushort) expected.Length);
+
+            Assert.Equal(0, mbbsEmuCpuRegisters.AX);
+            Assert.Equal(0, mbbsEmuCpuRegisters.DX);
```
paladine, line 42: this function turned out to be `void` right? I believe I was mistaken when I said it should return a value. If `void`, just don't assert on the return value at all.

paladine (changes requested):

MBBSEmu.Tests/ExportedModules/Majorbbs/movmem_Tests.cs
tuday2: Yes this is void

MBBSEmu.Tests/ExportedModules/Majorbbs/memset_Tests.cs
tuday2: memset is void as well, so I removed the asserts
```
        ///     Sets the first num bytes of the block of memory with the specified value
        ///
        ///     Signature: void memset(void *ptr, int value, size_t num);
```

tuday2:
Changed itoa ordinal to return the point and asserted on NotEqual 0

MBBSEmu.Tests/ExportedModules/Majorbbs/memset_Tests.cs
paladine: This comment is wrong inside our code. `string.h` has the actual definition, and it's `void _FAR * _RTLENTRYF _EXPFUNC memset(void _FAR *__s, int __c, size_t __n);`

**void _FAR * **

So it actually does return a pointer value, and mbbsemu is wrong

tuday2: Fixed.

tuday2:
Fixed memset to return value, updated documentation

MBBSEmu.Tests/ExportedModules/Majorbbs/itoa_Tests.cs
```diff
@@ -36,8 +36,8 @@ public void itoa_Test(ushort intConvert, ushort baseInt, string expected)
                 });
 
             //Verify Results
-            Assert.Equal(0, mbbsEmuCpuRegisters.AX);
-            Assert.Equal(0, mbbsEmuCpuRegisters.DX);
+            Assert.NotEqual(0, mbbsEmuCpuRegisters.AX);
```
paladine: for extra credit, assert that the pointer equals the one passed in

paladine (approved):

MBBSEmu.Tests/ExportedModules/Majorbbs/itoa_Tests.cs
paladine: Resolved in next commit

PR #310, opened 2020-11-22
Create strtol test, sortstgs test, fix typo Majorbbs.cs

tuday2:


tuday2:
@paladine All yours

MBBSEmu.Tests/ExportedModules/Majorbbs/sortstgs_Tests.cs
```diff
@@ -0,0 +1,64 @@
...
+            Reset();
+
+            //Set Argument Values to be Passed In
+            var stringPointer = new IntPtr16[inputArray.Length];
+
+            for (var i = 0; i < inputArray.Length; i++)
+                stringPointer[i] = mbbsEmuMemoryCore.AllocateVariable("INPUT_STRING"+ i, (ushort)(inputArray[i].Length + 1));
+
+            for (var i = 0; i < inputArray.Length; i++)
+                mbbsEmuMemoryCore.SetArray("INPUT_STRING"+ i, Encoding.ASCII.GetBytes(inputArray[i]));
```
paladine, line 31: Can move this statement inside the prior `for` loop

MBBSEmu.Tests/ExportedModules/Majorbbs/sortstgs_Tests.cs
```diff
@@ -0,0 +1,64 @@
...
+    public class sortstgs_Tests : ExportedModuleTestBase
+    {
+        private const int SORTSTGS_ORDINAL = 558;
+        
+        [Theory]
+        [InlineData(new[] { "This", "is", "an", "array" }, new[] { "an", "array", "is", "This" })]
+        [InlineData(new[] { "Hello", "my", "name", "is" }, new[] { "Hello", "is", "my", "name" })]
+        [InlineData(new[] { "Hello", "Anger" }, new[] { "Anger", "Hello" })]
+        [InlineData(new[] { "zzzz", "BBBB", "CCCC", "aaaa", "000AAA" }, new[] { "000AAA", "aaaa", "BBBB", "CCCC", "zzzz" })]
+        [InlineData(new[] { "zzzz", "BBBB", "CCCC", "aaaa", "000AAA", "111ZZZ", "!!!XXX" }, new[] { "!!!XXX", "000AAA", "111ZZZ", "aaaa", "BBBB", "CCCC", "zzzz" })]
```
paladine, line 18: also add `[InlineData(new string[] {}, new string[] {});`

MBBSEmu.Tests/ExportedModules/Majorbbs/sortstgs_Tests.cs
```diff
@@ -0,0 +1,64 @@
...
+        public void sortstgs_Test(string[] inputArray, string[] expectedArray)
+        {
+            //Reset State
+            Reset();
+
+            //Set Argument Values to be Passed In
+            var stringPointer = new IntPtr16[inputArray.Length];
+
+            for (var i = 0; i < inputArray.Length; i++)
+                stringPointer[i] = mbbsEmuMemoryCore.AllocateVariable("INPUT_STRING"+ i, (ushort)(inputArray[i].Length + 1));
```
paladine, line 28: nit: add extra space after the string and before the `+` sign, i.e.

`"INPUT_STRING" + i`

MBBSEmu.Tests/ExportedModules/Majorbbs/sortstgs_Tests.cs
```diff
@@ -0,0 +1,64 @@
...
+
+            //Return Values
+            var resultPointer = new IntPtr16[inputArray.Length];
+            var resultString = new string[inputArray.Length];
+
+            for (var i = 0; i < inputArray.Length; i++)
+                resultPointer[i] = mbbsEmuMemoryCore.GetPointer(arrayPointer + (i * IntPtr16.Size));
+
+            for (var i = 0; i < inputArray.Length; i++)
+                resultString[i] = Encoding.ASCII.GetString(mbbsEmuMemoryCore.GetString(resultPointer[i]));
```
paladine, line 56: ditto, can be combined with the above `for` loop as well

MBBSEmu.Tests/ExportedModules/Majorbbs/strtol_Tests.cs
```diff
@@ -0,0 +1,34 @@
...
+{
+    public class strtol_Tests : ExportedModuleTestBase
+    {
+        private const int STRTOL_ORDINAL = 586;
+
+        [Theory]
+        [InlineData("123456 Hello some text", 7, 10, 123456)]
+        [InlineData("2345678 YESSS", 7, 10, 2345678)]
+        [InlineData("78 A", 7, 10, 78)]
+        [InlineData("0x23CACE A very big number", 1, 16, 2345678)]
```
paladine, line 15: let's do a test with an empty string as well as a string that isn't actually a number, i.e. all text/non-digit

paladine, line 15: i'm pretty sure `strtol` can't handle the `0x` prefix on hex digits. It would return 0 in this case since x stops the parsing. Let me verify though and I'll chime back in on this thread.

paladine: And a negative number

paladine: Scratch that, the docs indicate `0x` is valid, at least for radix 16. 

Also of note is that this function accepts whitespace up front, as well as leading `-` or `+` signs, so add some tests that cover those cases as well

MBBSEmu.Tests/ExportedModules/Majorbbs/strtol_Tests.cs
```diff
@@ -0,0 +1,34 @@
...
+using Xunit;
+
+namespace MBBSEmu.Tests.ExportedModules.Majorbbs
+{
+    public class strtol_Tests : ExportedModuleTestBase
+    {
+        private const int STRTOL_ORDINAL = 586;
+
+        [Theory]
+        [InlineData("123456 Hello some text", 7, 10, 123456)]
```
paladine, line 12: how is `ushortPtr` being used here? We should be asserting on the returned pointer that it points to the proper index. I guess for the first value of "123456", we should assert that the returned pointer is src + 6, since it reads 6 characters and should point the end to the space between "6" and "Hello"

The next test should assert on a value of 7, "78" would be 2 ahead of course.

paladine (changes requested):

MBBSEmu.Tests/ExportedModules/Majorbbs/strtol_Tests.cs
paladine: We'll also want to have a `[Fact]` test that sets the `endptr` to `null` to verify it doesn't try to write to the `null` value.

paladine: After looking at the `endptr` stuff is implemented, I'm pretty sure it's not implemented correctly. The tests will validate this pretty easily.

tuday2: @paladine There is a + 1 in the ordinal that is making it 1 off from your logic above -- before I remove can you take a look now? 

I have it passing all tests and also properly getting the pointer afterwards. I also added the NULL fact test and added a couple more things to the ordinal to catch the leading blanks, blank string, and no numbers at all.

Let me know if i should remove the +1 from the ordinal (see below where it sets the suffix pointer):

```
if (suffixPointer != IntPtr16.Empty)
                Module.Memory.SetPointer(suffixPointer,
                    new IntPtr16(stringPointer.Segment, (ushort)(stringPointer.Offset + longToParseLength + 1)));
```

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -6443,13 +6443,28 @@ private void strtol()
...
+                    Module.Memory.SetPointer(suffixPointer, new IntPtr16(stringPointer.Segment, stringPointer.Offset));
+                
+                return;
+            }
+
+            var longToParse = stringContainingLongs.Split(' ', StringSplitOptions.RemoveEmptyEntries)[0];
+            var longToParseLength = longToParse.Length + (stringContainingLongs.Length - stringContainingLongs.TrimStart(' ').Length); //We do this as length might change with logic below and add back in leading spaces
 
-            if (longToParseLength == 0)
+            if (longToParseLength == 0 || !longToParse.Any(char.IsDigit))
```
paladine, line 6460: `char.IsDigit` likely isn't safe since hex values can have A-F for example

paladine (approved):

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
tuday2: I added a test for 0xFFFFF and also a max value test

tuday2:
Fixed digit check to only check on Radix == 10

paladine (approved):

PR #311, opened 2020-11-26
Fix extra chars in buffer from leaking into module

tuday2:
*Set InputBuffer length to 1 before entering module
*Fixes extra characters from being sent to module (when menu items >9 or menuoption key that is 2 chars)
*Modules behave strangely when setting to 0


tuday2:
@enusbaum Changed approach -- restricted menu item to 1 character

moved logic from majorbbs.cs to initial module load to avoid duplicating existing module MenuOptionKeys with auto generated ones

enusbaum (approved):

PR #315, opened 2020-12-08
EXE Loader

enusbaum:
- Added the ability to run an EXE file using the `-exe` Command Line switch
- Moved Interrupt Handlers to their own class w/ Interface of `InterruptHandler`
- Implemented Handlers for INT `21h` and `1Ah`
- Added Loading of MZ EXE Files (Headers + Program Loading)
- Several new Opcodes for EXE Support and enhanced existing ones


enusbaum:
Currently runs and executed a simple C program compiled with Turbo C:

```c
#include <stdio.h>

int main()
{
   printf("Hello World!");
   return 0;
}
```

Prints the result using INT `21h` `0x40` and exits with code 0

MBBSEmu/CPU/CPUCore.cs
```diff
@@ -2974,6 +2914,18 @@ private void Op_Scasb()
...
                     Registers.DI++;
                 }
+
+                if (_currentInstruction.HasRepnePrefix && Registers.CX > 0 && result != 0)
+                {
+                    Registers.CX--;
+                    goto scasb;
+                }
+
+                if (_currentInstruction.HasRepePrefix && Registers.CX > 0 && result == 0)
```
paladine, line 2924: is this necessary - two identical if blocks with result != 0 && result == 0

combine into single `if (_currentInstruction.HasRepePrefix && Registers.CX > 0)` ?

MBBSEmu/CPU/CPUCore.cs
```diff
@@ -1674,6 +1593,11 @@ private void Op_Retf()
         private void Op_Ret()
         {
             Registers.IP = Pop();
+
+            //Pop N bytes (N/2 words) that were Pushed before the CALL
+            if (_currentInstruction.Op0Kind == OpKind.Immediate16)
+                for (var i = 0; i < GetOperandValueUInt16(OpKind.Immediate16, EnumOperandType.Destination) / 2; i++)
+                    Pop();
```
paladine, line 1600: instead of looping through this, can't you just increment SP by the value and call it a day?

MBBSEmu/DOS/ExeRuntime.cs
```diff
@@ -0,0 +1,142 @@
...
+        {
+            Memory.AddSegment(ENVIRONMENT_SEGMENT);
+
+            if (_environmentVariables == null)
+                return;
+
+            ushort bytesWritten = 0;
+            foreach (var v in _environmentVariables)
+            {
+                Memory.SetArray(ENVIRONMENT_SEGMENT, bytesWritten, Encoding.ASCII.GetBytes(v));
```
paladine, line 137: does this add the trailing null? (i think I see now, probably get the null from the initial ZeroMemory allocation)

I believe this structure is doubly null terminated as well

someValue=something\0
another=whatever\0
\0\0

MBBSEmu/DOS/ExeRuntime.cs
```diff
@@ -0,0 +1,142 @@
...
+        /// <summary>
+        ///     Segment which contains the PSP Struct
+        ///
+        ///     Normally +10h beyond the last segment of the EXE
+        /// </summary>
+        private readonly ushort PSP_SEGMENT;
+
+        private readonly List<string> _environmentVariables;
+
+        public ExeRuntime(MZFile file, List<string> environmentVariables, ILogger logger)
```
paladine, line 37: Would environmentVariables be better as a `Dictionary<string, string>` with key/value? I'm assuming right now it's in the form `someVariable=someValue` for each item

MBBSEmu/DOS/ExeRuntime.cs
```diff
@@ -0,0 +1,142 @@
...
+        private const ushort ENVIRONMENT_SEGMENT = 0xFFF0;
+
+        /// <summary>
+        ///     Segment which contains the PSP Struct
+        ///
+        ///     Normally +10h beyond the last segment of the EXE
+        /// </summary>
+        private readonly ushort PSP_SEGMENT;
+
+        private readonly List<string> _environmentVariables;
```
paladine, line 35: perfect chance for an init-only property :)

MBBSEmu/DOS/Interrupts/IInterruptHandler.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace MBBSEmu.DOS.Interrupts
+{
+    public interface IInterruptHandler
+    {
+        ushort Vector { get; }
```
paladine, line 5: aren't there only 256 max?

MBBSEmu/DOS/Interrupts/Int1Ah.cs
```diff
@@ -0,0 +1,55 @@
...
+using MBBSEmu.Memory;
+
+namespace MBBSEmu.DOS.Interrupts
+{
+    /// <summary>
+    ///     Interrupt Vector 1Ah which handles BIOS calls for Clock Information
+    /// </summary>
+    public class Int1Ah : IInterruptHandler
+    {
+        private readonly CpuRegisters _registers;
```
paladine, line 14: moar init only setters? perhaps a trial run

MBBSEmu/DOS/Interrupts/Int1Ah.cs
```diff
@@ -0,0 +1,55 @@
...
+        public void Handle()
+        {
+            switch (_registers.AH)
+            {
+                case 0x0:
+                {
+                        /*
+                            INT 1A - AH = 00h CLOCK - GET TIME OF DAY
+                            Return: CX:DX = clock count
+                            AL = 0 if clock was read or written (via AH=0,1)
```
paladine, line 34: we need to compute AL right?

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -0,0 +1,280 @@
...
+                    }
+                case 0x25:
+                    {
+                        /*
+                            INT 21 - AH = 25h DOS - SET INTERRUPT VECTOR
+                            AL = interrupt number
+                            DS:DX = new vector to be used for specified interrupt
+                         */
+
+                        //TODO -- Implement, ignore for now
```
paladine, line 64: this seems implement, no more need for this comment?

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -0,0 +1,280 @@
...
+                    }
+                case 0x2F:
+                    {
+                        //Get DTA address
+                        /*
+                         *  Action:	Returns the segment:offset of the current DTA for read/write operations.
+                            On entry:	AH = 2Fh
+                            Returns:	ES:BX = Segment.offset of current DTA
+                         */
+                        if (DiskTransferArea == null && !_memory.TryGetVariablePointer("Int21h-DTA", out DiskTransferArea))
```
paladine, line 91: use GetOrAllocateVariablePointer instead?

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -0,0 +1,280 @@
...
+                        /*
+                             DOS Default/Predefined Handles:
+                             0 - Standard Input Device - can be redirected (STDIN)
+	                         1 - Standard Output Device - can be redirected (STDOUT)
+	                         2 - Standard Error Device - can be redirected (STDERR)
+	                         3 - Standard Auxiliary Device (STDAUX)
+	                         4 - Standard Printer Device (STDPRN)
+                         */
+
+                        if (fileHandle <= 2)
```
paladine, line 167: shouldn't write to fileHandle 0 (STDIN)

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -0,0 +1,280 @@
...
+                    {
+                        /*
+                            DOS 2+ - GET CURRENT DIRECTORY
+                            DL = drive (0=default, 1=A, etc.)
+                            DS:DI points to 64-byte buffer area
+                            Return: CF set on error
+                            AX = error code
+                            Note: the returned path does not include the initial backslash
+                         */
+                        _memory.SetArray(_registers.DS, _registers.SI, Encoding.ASCII.GetBytes("BBSV6\\\0"));
```
paladine, line 221: Add a TODO here to implement this properly?

MBBSEmu/Disassembler/MZFile.cs
```diff
@@ -0,0 +1,67 @@
...
+            //Get First Instruction
+            var dataSegmentStart = BitConverter.ToUInt16(programData.ToArray(), RelocationRecords[0].Offset);
+            dataSegmentStart <<= 4;
+
+            Segments.Add(new Segment() { Data = programData.Slice(0, dataSegmentStart).ToArray(), Flag = (ushort)EnumSegmentFlags.Code, Ordinal = 1 });
+            Segments.Add(new Segment() { Data = programData.Slice(dataSegmentStart).ToArray(), Flag = (ushort)EnumSegmentFlags.Data, Ordinal = 2 });
+        }
+
+        /// <summary>
+        ///     Starting at the
```
paladine, line 50: don't leave me hanging!

MBBSEmu/Memory/IMemoryCore.cs
```diff
@@ -293,6 +293,23 @@ public interface IMemoryCore
...
+        /// <param name="value"></param>
+        void SetArray(ushort segment, ushort offset, ushort count, byte value);
+
+        /// <summary>
+        ///     Writes the specified byte the specified number of times starting at the specified pointer
+        /// </summary>
+        /// <param name="pointer"></param>
+        /// <param name="count"></param>
+        /// <param name="value"></param>
+        void SetArray(IntPtr16 pointer, ushort count, byte value);
```
paladine, line 311: I think this would be more aptly named with `FillArray`, what do you think?

MBBSEmu/Memory/MemoryCore.cs
```diff
@@ -506,6 +506,28 @@ public void SetArray(ushort segment, ushort offset, ReadOnlySpan<byte> array)
...
+        /// <summary>
+        ///     Writes the specified byte the specified number of times starting at the specified pointer
+        /// </summary>
+        /// <param name="segment"></param>
+        /// <param name="offset"></param>
+        /// <param name="count"></param>
+        /// <param name="value"></param>
+        public void SetArray(ushort segment, ushort offset, ushort count, byte value)
+        {
+            for (var i = 0; i < count; i++)
```
paladine, line 518: Can probably use some fancy `Span` methods to achieve this I would think

paladine (changes requested):

MBBSEmu/CPU/CPUCore.cs
enusbaum: They're not identical. One is for 'REPE' prefix (continue looping if results are equal) and 'REPNE' prefix (continue looping if results are not equal)

MBBSEmu/DOS/ExeRuntime.cs
enusbaum: They're null terminated in memory so I handled them as they're handled by DOS as an array of strings. Since they're written to the Environment Variable sector as `key=value\0` strings, there's not much value in splitting them.

MBBSEmu/DOS/ExeRuntime.cs
enusbaum: The line below I increment length +1 since the array is initialized with `0x0` values. Since it's writing to a dedicated sector, it'll have trailing nulls after the final result so the double-null shouldn't be an issue.

MBBSEmu/DOS/Interrupts/IInterruptHandler.cs
enusbaum: Yep -- good catch, I'm pulling the 'Immediate8' off the opcode for `INT`. I'll update the signature.

MBBSEmu/DOS/Interrupts/Int21h.cs
enusbaum: It's not wrong -- technically. All MBBS Modules ran within the context of the same directory as MajorBBS (`BBSV6` in most cases), so this wouldn't change as addon utilities also ran within the same folder.

MBBSEmu/Disassembler/MZFile.cs
enusbaum: 🤣😆🤣😆🤣😆

MBBSEmu/Memory/IMemoryCore.cs
enusbaum: I was tying to be consistent with the `Set` taxonomy (`SetByte`, `SetWord`, `SetArray`). But renaming it would keep the overloads simpler.

MBBSEmu/Memory/MemoryCore.cs
enusbaum: Resolved using `Array.FIll()`

MBBSEmu.CPU.Benchmark/MBBSEmu.CPU.Benchmark.csproj
```diff
@@ -3,6 +3,7 @@
   <PropertyGroup>
     <OutputType>Exe</OutputType>
     <TargetFramework>net5.0</TargetFramework>
+    <LangVersion>preview</LangVersion>
```
paladine, line 6: 9.0?

paladine (approved):

MBBSEmu/DOS/ExeRuntime.cs
paladine: Works for me

MBBSEmu/DOS/ExeRuntime.cs
paladine: Might be useful for later if we grab all the env variables from the C# program and pass them to the running EXE. Either form works, just might require more conversion down the line.

MBBSEmu/DOS/Interrupts/Int21h.cs
paladine: Should be OK for now, though it might impact actual file operations down the line. Not pressing at this time at least

MBBSEmu.CPU.Benchmark/MBBSEmu.CPU.Benchmark.csproj
enusbaum: Whoops! Got a little YOLO there

paladine (approved):

PR #321, opened 2020-12-12
Add IP2LOCATION support

tuday2:
-Add config entry to limit telnet connections to a country based on IP address
-Check upon connection if the IP is from an allowed country
-Fixed typo in rlogin config

TODO:
-Add multiple country support
-Add config error checking

NOTES:
https://lite.ip2location.com/ - website to get updated IP2LOCATION country list , updated every month
Checked FAQ and the "lite" version (which this is based on) is able to be distributed freely
(I got the idea from Mystic BBS -- they support this)


tuday2:
-Changed IP2LOCATION to use API, no more big files or messy loads
-demo key can lookup 20 per day and you can get a free trial key for 5000 lookups -- fails open if any issues with API
-Unless enabled completely ignores logic

PR #324, opened 2020-12-12
Change Sysop command to /SYS

tuday2:
prevent interference with modules


enusbaum:
Need to also change all of the help text

enusbaum (approved):

PR #335, opened 2020-12-21
Condex() test and sysop version

tuday2:


paladine (approved):
Fine by me. Though in the future probably best to split this into two separate PRs to better isolate each independent change. Plus it doubles your PR stats ;-)

PR #344, opened 2020-12-24
Fix anpbtvlk

paladine:
Figured out what it's supposed to do, which helps

### DIFF
--- a/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
+++ b/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
@@ -1287,7 +1287,6 @@ namespace MBBSEmu.HostProcess.ExportedModules
             if (bb.IsNull())
                 return false;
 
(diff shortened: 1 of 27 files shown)


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -1281,19 +1282,21 @@ private void anpbtvl()
...
         {
-            var currentBtrieveFile = Module.Memory.GetPointer("BB");
-            if (currentBtrieveFile.IsNull())
-            {
+            var bb = Module.Memory.GetPointer("BB");
+            if (bb.IsNull())
                 return false;
-            }
+
+            var btrieveFileProcessor = BtrieveGetProcessor(bb);
```
enusbaum, line 1289: We can remove this, not used

enusbaum (changes requested):

enusbaum (approved):

PR #353, opened 2020-12-27
Limit primary processing core to only run at set intervals = CPU REDUCTION

paladine:
Adds a wait to the main loop. The wait runs at a configurable hertz, controlled in `Timer.Hertz`

Value of 0 is full speed, or mbbsemu as it is today. Value of 1 is once a second, 2 is twice a second, etc. 

The code is smart enough to abort the waiting when user input is received. The socket threads will start the main loop and process accordingly, so it'll seem like a real time interaction without delay/jitter. 

Given that rti handlers run at 18 hz, most users should probably set this to 18 at a minimum, but lower values can work depending on which modules are loaded.

### DIFF
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -90,6 +90,7 @@ namespace MBBSEmu.HostProcess
         ///     Timer that triggers when nightly cleanup should occur
         /// </summary>
         private readonly Timer _timer;
+        private readonly Timer _tickTimer;
 
         /// <summary>
         ///     Flag that controls whether the main loop will perform a nightly cleanup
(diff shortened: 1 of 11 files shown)


MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -129,6 +135,13 @@ public MbbsHost(IClock clock, ILogger logger, IGlobalCache globalCache, IFileUti
             _incomingSessions = new Queue<SessionBase>();
             _cleanupTime = _configuration.CleanupTime;
             _timer = new Timer(unused => _performCleanup = true, this, NowUntil(_cleanupTime), TimeSpan.FromDays(1));
+
+            if (_configuration.TimerHertz > 0)
+            {
+                _timerEvent = new AutoResetEvent(true);
+                _tickTimer = new Timer(unused => _timerEvent.Set(), this, TimeSpan.Zero, TimeSpan.FromMilliseconds(1000 / configuration.TimerHertz));
```
enusbaum, line 142: you can use `_` for a discard parameter vs `unused`. won't allocate

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -183,6 +200,17 @@ public void WaitForShutdown()
             _workerThread.Join();
         }
 
+        private void WaitForNextTick()
+        {
+            if (_timerEvent == null ||
+                _channelDictionary.Values.Where(session => session.DataFromClient.Count > 0 || session.DataToClient.Count > 0 || session.DataToProcess).Any())
```
enusbaum, line 206: You can do `.Any()` in place of the `.Where()`

enusbaum (changes requested):
Finished review of the code, I'll run it through a couple modules to test things out.

enusbaum:
@tuday2 Ran it through the paces with multiple modules and we look good!

PR #358, opened 2020-12-30
Add protection around values of Timer.Hertz.

paladine:
0-1000 is valid, 36 is default. Above 1000 goes back to 0


tuday2 (approved):
Thanks for changing to Logger

PR #359, opened 2020-12-30
Support for MajorBBS/Worldgroup Modules with Multiple DLL's (Farwest Trivia)

enusbaum:
- Added support for an `MbbsModule` object to have multiple `ModuleDlls`
- All DLL's required for the module are loaded and relocations patched in the same `IMemoryCore`
- The "main" Module DLL is always loaded at `ModuleDlls[0]`
- `_INIT_` routine for all Module DLL's is invoked, in alphabetical order (as it seems it is on MajorBBS)
- PHAPI init (?) code starting at `0001:0000` is now patched with a `retf`, skipping the code entirely (as I don't believe it's needed)
- Cleaned up how RTKICK, RTIHDLR, TASKS, etc. routines are saved & executed
- Added 32-Bit General Purpose Registers (EAX, EBX, ECX, EDX) to the Registers class along with updated Helpers
- Extended several x86 opcodes to support 32-bit operations (XOR, SUB, SHLD, ADD, MUL, etc.) and added Unit Tests for the new 32-bit scenarios
- Refactored `IInterruptHandler` for `Int21h` to require the currently executing path be sent into it (for Interrupt based file I/O operations)
- Extended `IMemoryCore` to include double word operations

image

Farwest Trivia now loads and works -- but there are some more missing ordinals. We can address them later.


MBBSEmu.Tests/CPU/INC_Tests.cs
```diff
@@ -0,0 +1,178 @@
...
+namespace MBBSEmu.Tests.CPU
+{
+    public class INC_Tests : CpuTestBase
+    {
+        [Theory]
+        [InlineData(0,1, false, false, false)]
+        [InlineData(0xFF, 0, true, false, false)]
+        [InlineData(0x7F, 0x80, false, true, true)]
+        [InlineData(0x80, 0x81, false, false, true)]
+        public void INC_R8(byte alValue, byte expectedResult, bool zeroFlagSet, bool overflowFlagSet, bool signFlagSet)
```
paladine, line 16: I feel like a lot of these tests could be more concise if instead of passing 3-4 flag values in the `InlineData` areas, you just pass a single expected `EnumFlags` with a combination of all the expected things.

MBBSEmu.Tests/ExportedModules/Majorbbs/register_module_Tests.cs
```diff
@@ -35,21 +35,21 @@ public void register_module_test()
...
-            Assert.Equal(new FarPtr(0x0002, 0x0002), mbbsModule.EntryPoints["dlarou"]);
-            Assert.Equal(new FarPtr(0x0003, 0x0003), mbbsModule.EntryPoints["finrou"]);
-            Assert.Equal(new FarPtr(0x0004, 0x0004), mbbsModule.EntryPoints["huprou"]);
-            Assert.Equal(new FarPtr(0x0005, 0x0005), mbbsModule.EntryPoints["injrou"]);
-            Assert.Equal(new FarPtr(0x0006, 0x0006), mbbsModule.EntryPoints["lofrou"]);
-            Assert.Equal(new FarPtr(0x0007, 0x0007), mbbsModule.EntryPoints["lonrou"]);
-            Assert.Equal(new FarPtr(0x0008, 0x0008), mbbsModule.EntryPoints["mcurou"]);
-            Assert.Equal(new FarPtr(0x0009, 0x0009), mbbsModule.EntryPoints["stsrou"]);
-            Assert.NotEqual(FarPtr.Empty, mbbsEmuMemoryCore.GetPointer(mbbsEmuMemoryCore.GetVariablePointer("MODULE") + (2 * mbbsModule.StateCode)));
+            Assert.Equal(new FarPtr(0x0001, 0x0001), mbbsModule.ModuleDlls[0].EntryPoints["sttrou"]);
```
paladine, line 38: a property such as `PrimaryModuleDll` or `MainModuleDll` would be useful here

MBBSEmu.Tests/ExportedModules/Majorbbs/stzcat_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+    public class stzcat_Tests : ExportedModuleTestBase
+    {
+        private const int STZCAT_ORDINAL = 905;
+
+        [Theory]
+        [InlineData("12345", "12345", 7, "123451")]
+        [InlineData("12345", "12345", 4, "123")]
+        [InlineData("12345", "12345", 0, "")]
+        [InlineData("12345", "12345", 11, "1234512345")]
+        [InlineData("", "", 0, "")]
```
paladine, line 20: we need a test for a large buffer, to ensure the zero fill works as expected

Like `[InlineData("12345", "12345", 32, "1234512345")]` and verify the zero padding at the end (also initialize all of `dst` with something != 0)

MBBSEmu/CPU/CPUCore.cs
```diff
@@ -3437,6 +3663,54 @@ private void Op_Fchs()
...
+        }
+
+        private uint Op_Shld_32()
+        {
+            var destination = GetOperandValueUInt32(_currentInstruction.Op0Kind, EnumOperandType.Destination);
+            var source = GetOperandValueUInt32(_currentInstruction.Op1Kind, EnumOperandType.Source);
+            var count = GetOperandValueUInt8(_currentInstruction.Op2Kind, EnumOperandType.Count);
+
+            //Make sure Count isn't > 32 bits
+            if (count >= sizeof(uint) * 8)
```
paladine, line 3691: prior code had `count >` and this is `count >=`, just wanted to clarify if this is accurate

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -268,6 +279,43 @@ 16h DEC
...
+                                         bytes 18h-19h: file date
+                                         bytes 1Ah-1Dh: file size
+                                         bytes 1Eh-3Ah: ASCIZ filename+extension
+                         */
+                        var fileName = Encoding.ASCII.GetString(_memory.GetString(_registers.DS, _registers.DX, true));
+
+                        var fileUtility = new FileUtility(_logger);
+                        var foundFile = fileUtility.FindFile(_path, fileName);
+
+                        if(!File.Exists($"{_path}{fileName}"))
```
paladine, line 314: this should check against `foundFile` no?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -1459,6 +1463,40 @@ private void stzcpy()
...
+        /// <summary>
+        ///     Concatenates Two Strings with a fixed length
+        ///
+        ///     Signature: char *stzcat(char *dst,char *src,int num);
+        /// </summary>
+        private void stzcat()
+        {
+            var destinationPointer = GetParameterPointer(0);
+            var sourcePointer = GetParameterPointer(2);
+            var limit = GetParameter(4);
```
paladine, line 1475: more accurately `dstBufferLength` or `dstLength`

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -1459,6 +1463,40 @@ private void stzcpy()
...
+
+            var stringLength = Module.Memory.GetString(destinationPointer, true).Length;
+
+            //Truncate Destination if LIMIT is less than DST 
+            if (stringLength > limit)
```
paladine, line 1480: would this ever happen, the string being longer than the total buffer length? That's a pretty major module bug hah

paladine (changes requested):

MBBSEmu.Tests/CPU/INC_Tests.cs
```diff
@@ -0,0 +1,178 @@
...
+using Xunit;
+using static Iced.Intel.AssemblerRegisters;
+
+namespace MBBSEmu.Tests.CPU
+{
+    public class INC_Tests : CpuTestBase
+    {
+        [Theory]
+        [InlineData(0,1, false, false, false)]
+        [InlineData(0xFF, 0, true, false, false)]
```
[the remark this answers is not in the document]
enusbaum: I verified using NASM and it doesn't set overflow on AL.

image

MBBSEmu.Tests/ExportedModules/Majorbbs/register_module_Tests.cs
enusbaum: Added property `MainModuleDll` to `MbbsModule` and updated references to `[0]` to make them easier to understand

MBBSEmu/CPU/CPUCore.cs
enusbaum: Updated this to a modulo 32 as per the documentation. The previous IF was for the case listed where if the shift amount was `>=` the operand size.

image

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: I mean... it shouldn't....but it could.

MBBSEmu.Tests/CPU/INC_Tests.cs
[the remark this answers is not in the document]
paladine: Testing with asm, impressive, most impressive. This article claims that OF is set when an inc of 127 occurs at least: https://stackoverflow.com/questions/3925528/x86-assembly-inc-and-dec-instruction-and-overflow-flag. 

We can probably update/validate this in another PR

paladine (approved):

PR #362, opened 2021-01-01
Add configurable logging level and file logging

tuday2:
-Started cleanup of all Logger calls
-Added module identifier where possible
-Started with console, do we want to add file? seems like you have it ready just need to add config

Wanted to get feedback on any changes/formatting/etc. before I get too much farther

### DIFF
--- a/MBBSEmu/Configuration.cs
+++ b/MBBSEmu/Configuration.cs
@@ -7,6 +7,7 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Text.Json;
+using MBBSEmu.Logging;
 
 namespace MBBSEmu
 {
@@ -58,9 +59,9 @@ namespace MBBSEmu
         public string ANSILogoff => ConfigurationRoot["ANSI.Logoff"];
         public string ANSISignup => ConfigurationRoot["ANSI.Signup"];
         public string ANSIMenu => ConfigurationRoot["ANSI.Menu"];
-        public string ConsoleLogLevel => GetLogLevel("Console.LogLevel");
+        public string ConsoleLogLevel => ConfigurationRoot["Console.LogLevel"];
         public string FileLogName => GetFileNameAppSettings("File.LogName");
-        public string FileLogLevel => GetLogLevel("File.LogLevel");
+        public string FileLogLevel => ConfigurationRoot["File.LogLevel"];
         public IEnumerable<string> DefaultKeys
         {
             get
@@ -179,25 +180,6 @@ namespace MBBSEmu
                 timerHertz = 36;
             }
             return timerHertz;
-        }
-
-        private string GetLogLevel(string key)
-        {
-            var logLevel = ConfigurationRoot[key];
-            var logLevels = new[] {"Debug", "Info", "Warn", "Error", "Fatal"};
-
-            if (string.IsNullOrEmpty(logLevel))
-            {
-                _logger.Warn($"{key} not specified in {Program._settingsFileName ?? Program.DefaultEmuSettingsFilename} -- setting default value: Info");
-                logLevel = "Info";
-            }
-
-            if (logLevels.Any(x => logLevel.Equals(x, StringComparison.InvariantCultureIgnoreCase)))
-                return logLevel;
-            _logger.Warn($"{key} not recognized as valid logging level (Debug, Info, Warn, Error, Fatal) -- setting default value: Info");
-            logLevel = "Info";
-
-            return logLevel;
         }
 
         private string GetRemoteIPAppSettings(string key)
--- a/MBBSEmu/Configuration.cs
+++ b/MBBSEmu/Configuration.cs
@@ -157,7 +157,7 @@ namespace MBBSEmu
                     pathFile = "mbbsemu.db";
                     return pathFile;
                 case "File.LogName":
-                    _logger.Info($"No valid filename for logging, disabling");
+                    //File Logging Disabled
                     pathFile = "";
                     return pathFile;
                 default:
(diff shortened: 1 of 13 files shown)


MBBSEmu/Logging/CustomLogger.cs
```diff
@@ -14,20 +14,12 @@ public class CustomLogger : Logger
...
-                FileName = @"c:\dos\log\log.txt",
-                Layout = Layout.FromString("${shortdate} ${time} ${level} ${callsite} ${message}"),
-                DeleteOldFileOnStartup = true
-            };
-            //config.AddTarget(fileLogger);
+            config.AddTarget(consoleLogger);
             config.AddRuleForAllLevels(consoleLogger);
-            //config.AddRuleForAllLevels(fileLogger);
             LogManager.Configuration = config;
+            LogManager.Configuration.Variables["mbbsdir"] = Directory.GetCurrentDirectory() + "\\";
```
enusbaum, line 22: Use `Path.DirectorySeparatorChar` vs `\\` as this would not work on Linux/OSX

https://docs.microsoft.com/en-us/dotnet/api/system.io.path.directoryseparatorchar?view=net-5.0

MBBSEmu/Program.cs
```diff
@@ -234,6 +236,23 @@ private void Run(string[] args)
...
+                LogManager.Configuration.LoggingRules.Clear();
+                LogManager.Configuration.AddRule(LogLevel.FromString(configuration.ConsoleLogLevel), LogLevel.Fatal, "consoleLogger");
+                if(!string.IsNullOrEmpty(configuration.FileLogName))
+                {
+                    var fileLogger = new FileTarget("fileLogger")
+                    {
+                        FileNameKind = 0,
+                        FileName = "${var:mbbsdir}" + configuration.FileLogName,
+                        Layout = Layout.FromString("${shortdate} ${time} ${level} ${callsite} ${message}"),
+                        DeleteOldFileOnStartup = true
```
enusbaum, line 249: Do we want to make this optional/configurable/append by default? If someone crashed and restarted they'd lose the crash log.

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -1681,7 +1681,7 @@ private void stgopt()
 
 #if DEBUG
                 _logger.Info(
```
enusbaum, line 1683: We can probably make this `Debug` -- I don't think they'd be needed by an ordinary use at `Info` level.

MBBSEmu/Configuration.cs
```diff
@@ -168,6 +181,25 @@ private int GetTimerHertz(string key)
             return timerHertz;
         }
 
+        private string GetLogLevel(string key)
+        {
+            var logLevel = ConfigurationRoot[key];
+            var logLevels = new[] {"Debug", "Info", "Warn", "Error", "Fatal"};
```
enusbaum, line 187: We can probably use the nLog built in `LogLevel.FromString()` vs. defining them here.

https://nlog-project.org/documentation/v4.0.0/html/M_NLog_LogLevel_FromString.htm

MBBSEmu/Configuration.cs
```diff
@@ -139,17 +142,27 @@ public T GetAppSettings<T>(object value, string valueName)
...
-                {
-                    "Database.File" => new Exception($"Please set a valid database filename(eg: mbbsemu.db) in the {Program._settingsFileName ?? Program.DefaultEmuSettingsFilename} file before running MBBSEmu"),
-                    _ => new Exception($"Missing {key} in {Program._settingsFileName ?? Program.DefaultEmuSettingsFilename}"),
-                };
+                case "Database.File":
+                    _logger.Warn($"No valid database filename set in {Program._settingsFileName ?? Program.DefaultEmuSettingsFilename} -- setting default value: mbbsemu.db");
+                    pathFile = "mbbsemu.db";
+                    return pathFile;
+                case "File.LogName":
+                    _logger.Info($"No valid filename for logging, disabling");
```
enusbaum, line 160: I'm torn if we'll need this message. For example, I'll have it off forever because file logging is synchronous and murders my local performance when debugging.

enusbaum (changes requested):

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: And all the `opt` ordinals for that matter.

MBBSEmu/Configuration.cs
tuday2: It's not very clear how tolerant it is --- i am using it from take the config variable and put it into NLOG but here was trying to make sure it was a valid log level as to not crash anything

Program.cs:
```
LogManager.Configuration.AddRule(LogLevel.FromString(configuration.FileLogLevel), LogLevel.Fatal, "fileLogger");
```

MBBSEmu/Logging/CustomLogger.cs
paladine: Or better yet, `Path.Combine` or `Path.Join`

tuday2: I'm adding a filename on to the end of this and it needs a trailing slash. Join/Combine don't seem to address that.

MBBSEmu/Program.cs
tuday2: Removed this per our discussion

MBBSEmu/Configuration.cs
tuday2: Kept this as I think its the best way, I re-used it for both console and file

MBBSEmu/Configuration.cs
tuday2: Removed message, agree -- if someone wants file logging they will put it in there we don't need to warn them if they don't want it in the first place.

enusbaum (approved):

PR #371, opened 2021-01-10
Add samend() ordinal 813 + Tests

tuday2:
Needed for Othello

### DIFF
--- a/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
+++ b/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
@@ -3003,11 +3003,8 @@ namespace MBBSEmu.HostProcess.ExportedModules
         /// </summary>
         private void sameto()
         {
-            var string1Pointer = GetParameterPointer(0);
-            var string2Pointer = GetParameterPointer(2);
-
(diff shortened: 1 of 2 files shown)


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -3009,10 +3012,27 @@ private void sameto()
...
 
+        /// <summary>
+        ///     Compares ending of strings, checking if the first string ends with the second string (ignoring case)
+        ///
+        ///     Signature: int samend(char *longs, char *ends)
+        ///     Returns: AX == 1, match
+        /// </summary>
+        private void samend()
+        {
+            var string1Pointer = GetParameterPointer(0);
```
enusbaum, line 3023: You can use `GetParameterString` here and avoid the buffer declarations below

enusbaum (changes requested):

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
tuday2: Nice, fixed.

tuday2:
Fixed sameto() the same way since this just the "ends with" version, and retested.

enusbaum (approved):

PR #379, opened 2021-01-23
Add HostIP setting in AppSettings.json allowing you to Bind Host to a specific IP

tuday2:
-Defaults to "0.0.0.0" which is the same as IpAddress.Any
-If setting exists, checks to make sure that IP Address exists on the system (otherwise it gives a pretty cryptic error message about "no address found in this context"


enusbaum (changes requested):
I would split this into two options for `Rlogin.IP` and `Telnet.IP`, as there might be situations where Telnet is listening publicly and Rlogin is listening privately.

PR #383, opened 2021-01-26
Use a single Random from TestBase in tests

paladine:
Fixes flakiness? Should test with continual tests over a time period


paladine:
Doesn't seem to fix the flakiness, but doesn't hurt to merge anyhow?

PR #391, opened 2021-01-30
Fixed INT 21h Support

enusbaum:
Fixed INT 21h 0x47 (GetCurrentDirectory) while debugging RCIROSE. Verified the current implementation of these interrupts operates properly.

### DIFF
--- a/MBBSEmu/DOS/Structs/DTAStruct.cs
+++ b/MBBSEmu/DOS/Structs/DTAStruct.cs
@@ -1,0 +1,74 @@
+﻿using System;
+using System.Text;
+
+namespace MBBSEmu.DOS.Structs
+{
+    /// <summary>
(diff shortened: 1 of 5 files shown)


MBBSEmu/DOS/Structs/DTAStruct.cs
```diff
@@ -0,0 +1,74 @@
+﻿using System;
```
paladine, line 1: is this part of another branch? Doesn't seem to be used at all

paladine (approved):

MBBSEmu/DOS/Structs/DTAStruct.cs
enusbaum: It's a namespace within the .NET Framework, holds the `Array` class used for the data copying

PR #392, opened 2021-01-30
Add System variable replacement

tuday2:
-Tested adding a variable called DATE with BBSDRAW.EXE (MajorBBS ANSI program)


tuday2:
-Still need to add justification detection and padding (first two characters and last 2 numbers), sample below
R4TEXT_VAR_RJUS20
C9TEXT_VAR_CJ_25
N.TEXT_VAR_NOJUS

MBBSEmu/Session/SessionBase.cs
```diff
@@ -222,7 +225,14 @@ public void SendToClient(byte[] dataToSend)
         {
             if (OutputEnabled)
             {
-                SendToClientMethod(dataToSend.Where(shouldSendToClient).ToArray());
+                var dataToSendSpan = new ReadOnlySpan<byte>(dataToSend);
+                var sessionVariables = new Dictionary<string, string>
```
enusbaum, line 229: Can we set this in the SessionBase constructor so we're not newing it up with every output?

MBBSEmu/TextVariables/TextVariableService.cs
```diff
@@ -59,5 +64,105 @@ public int GetVariableIndex(string name)
...
+
+                //Get variable name
+                while (input[i] != 0x1)
+                {
+                    i++;
+                    variableNameLength++;
+                }
+
+                var variableName = Encoding.ASCII.GetString(input.Slice(variableNameStart, variableNameLength));
+                var variableText = GetVariableByName($"{variableName}");
```
enusbaum, line 112: I don't think we need the string formatter here, can just pass in `variableName`

enusbaum (changes requested):

enusbaum (approved):

PR #404, opened 2021-02-14
Add missing File Tests / Minor Tweaks

tuday2:
-Made File Pointer Dictionary Public for tests
-Changed fungetc() -> ungetc()
-Updated exceptions in 2 ordinals
-Changed fseek to take int instead of uint for Offset
-fputc()
-fgetc()
-fputs()
-fgets()
-fseek(0
-ungetc()

### DIFF
--- a/MBBSEmu.Tests/ExportedModules/Majorbbs/file_Tests.cs
+++ b/MBBSEmu.Tests/ExportedModules/Majorbbs/file_Tests.cs
@@ -1,3 +1,5 @@
+using MBBSEmu.Extensions;
+using MBBSEmu.HostProcess.Structs;
 using MBBSEmu.Memory;
 using System;
 using System.IO;
@@ -24,8 +26,6 @@ namespace MBBSEmu.Tests.ExportedModules.Majorbbs
         {
             Assert.Equal(LOREM_IPSUM_LENGTH, LOREM_IPSUM.Length);
         }
-
-        // TODO(add fgetc/fputc fgets/fputs fseek tests)
 
         [Theory]
         [InlineData("r")]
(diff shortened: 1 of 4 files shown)


MBBSEmu.Tests/ExportedModules/Majorbbs/file_Tests.cs
```diff
@@ -300,5 +300,329 @@ public void fclose_invalidPointer()
...
+            Assert.Equal(0, fseek(filep, 0, 0));
+
+            Assert.Equal(0, fseek(filep, fseekOffset, 1));
+
+            Assert.Equal(expectedChar, fgetc(filep));
+
+            Assert.Equal(0, fclose(filep));
+        }
+
+        [Theory]
```
paladine, line 430: could probably combine this with the test above it adding a third parameter to be passed to `fseek`

MBBSEmu.Tests/ExportedModules/Majorbbs/file_Tests.cs
```diff
@@ -300,5 +300,329 @@ public void fclose_invalidPointer()
...
+        }
+
+        [Fact]
+        public void fputs_InvalidStream_Throw()
+        {
+            //Reset State
+            Reset();
+
+            //Pass empty pointer
+            Assert.Throws<FileNotFoundException>(() => fputs(new FarPtr(), new FarPtr()));
```
paladine, line 605: technically this should return EOF and not throw :) But that'll require mods to majorbbs - would be a good follow on

paladine (approved):

MBBSEmu.Tests/ExportedModules/Majorbbs/file_Tests.cs
tuday2: Agree, will want to find a module that i can test with that acknowledges the filestream missing error -- will handle as a follow up

PR #407, opened 2021-02-19
Text/ASCII mode for fgets/fputs

synacktic:
Ok, so I made a rather obvious change to set the binary flag in the FileStruct. I don't think there is much to cover there.

As for how I handled fgets/fputs, let me explain the why.

1. Borland's fgets pulls from (f)getc. We do not currently make any of these functions call each other, but instead just have them all read the file stream. That being the case, I just made the changes within each function since the FileStream does not support a text mode. 
2. Borland's fgetc looks like it just strips all \r whether they are paired with a \n or not, so I did the same.
3. I just did a simple comparison, someone may want me to do this another way
4. If this all looks good for these two, I can update fgetc and whatever other functions might matter.

In short, after considering a million ways to handle this including using different stream types or messing with methods in FileStream, I decided that the simplest solution seemed to make the most sense based upon how things are done in other functions, and how it would not cause any ripple effects on functions that don't care about ascii vs text.


synacktic:
Oh and I used your cute little \r\n function (FormatNewLineCarriageReturn) for the output. Thanks for that.

enusbaum:
You'll need to add Unit Tests to `\MBBSEmu.Tests\ExportedModules\Majorbbs\file_Tests.csfile_Tests.cs` for the new case of ASCII mode. 

I'd suggest updating the test file contents with a `\r\n` and verifying the logic is correct.

synacktic:
k. Things explode, around seeking from the end being an obvious one because I changed the end of the test file, of course. Some of that is just that the expected results need updating because of the new offsets, but I'm going to create some test programs in dos with borland to make sure I understand correct seeking behavior before I go changing tests.

synacktic:
All right, based upon the test fails.. I went ahead and updated the fgetc function to handle text mode the way the DOS version does. It just returns \n in place of \r. If you seek to the last and second last characters and return them with fgetc you would get two \n instead of \r and \n. Verified this functionality by testing in DOS.

@paladine I think that should cover the currently implemented f* functions. 

I added a test to make sure fseek behaves with acsii opened files, but it really should not change it's behavior,  the offsets remain the same (also tested in DOS). So, I may want to just remove that one and add something that makes sure pulling the last could chars with fgetc does what is expected between binary and ascii and maybe that fgets behaves as expected.

synacktic:
Ok, the tests are now updated to make sure the newlines are handled correctly. Since text is default (I verified this is proper behavior), I also had to change the binary version of the test to say "rb" for its mode and not just "r".

I think both the testing changes and the ascii mode changes (ie everything) are good for review.

paladine (approved):

Issue #143, opened 2020-09-20
[MUICYBER] Lords of Cyberspace - Crashes during nightly cleanup


BlaZZZed1980:
**Module Information**
* MUICYBER
* Lords of Cyberspace

**Describe the bug**
Causes MBBSEmu to crash when performing nightly cleanup routines

**To Reproduce**
1. Load MUICYBER module
2. Wait for nightly cleanup to run

**Expected behavior**
Would like cleanup to run properly, or MBBSEmu to keep running after the error

**Log File Excerpt**
```
2020-09-19 03:01:25.2517 Info MBBSEmu.HostProcess.MbbsHost.<DoNightlyCleanup>b__44_1 Calling nightly cleanup routine on module MUICYBER
AUDIT SUMMARY: LoC: Clean Up Started 
AUDIT DETAIL: LoC: Clean Up Started 
AUDIT SUMMARY: LoC: Clearing Inactive Chars 
AUDIT DETAIL: LoC: Clearing Inactive Chars 
2020-09-19 03:01:25.2818 Warn MBBSEmu.HostProcess.ExportedModules.Majorbbs.f_read Attempting to read EOF file, returning null pointer
AUDIT SUMMARY: LoC: Checking Tournament 
AUDIT DETAIL: LoC: Checking Tournament 
AUDIT SUMMARY: LoC: Internet Billing 
AUDIT DETAIL: LoC: Internet Billing 
2020-09-19 03:01:25.2918 Warn MBBSEmu.HostProcess.ExportedModules.Majorbbs.f_open Reopened File: c:\modules\muicyber\MUICYBER\DATA\USERS.DAT - most likely a module bug.
```
**Software Information:**
- OS: Windows 7 32-bit VM
- MBBSEmu v1.0-alpha-083120

tuday2:
Adding the crash message:

2020-09-23 06:14:00.0371 Warn MBBSEmu.HostProcess.ExportedModules.Majorbbs.f_open Reopened File: D:\MbbsX64\modules\LRDCYBR\MUICYBER\DATA\USERS.DAT - most likely a module bug.
Unhandled exception. System.Exception: Unable to locate FileStream for 0106:B8DD (Stream: 0000:0000)
   at MBBSEmu.HostProcess.ExportedModules.Majorbbs.fseek()
   at MBBSEmu.HostProcess.ExportedModules.Majorbbs.Invoke(UInt16 ordinal, Boolean offsetsOnly)
[...]

PR #412, opened 2021-02-22
Multiple Open Perfection

synacktic:
I added a ref_count to the file struct that stops us from destroying it until it is the last usage. 

This was done to fix the double opens that were crashing Lords of Cyberspace per #143 . 

This is super basic, I would not be stunned if something could cause it to lose track or leave a bunch of memory in use.

That said, it does work and it does fix the cleanup for Lords of Cyberspace.

I won't be offended if you hate it, just let me know how it should be done. Let's call it a proof of concept. ;)


synacktic:
Quick note: one of the reasons this works is because the Filestream is r/w and I don't think we ever check the filestruct for the r/w flag, so if we reuse the "read" filestuct for writing, nothing complains.

paladine (changes requested):
changes look good, but we have no test coverage for this. Can you add a new test case in `file_tests.cs` which performs a double open and tries to operate on each simultaneously?

enusbaum (approved):

PR #419, opened 2021-02-28
Add setmode() and gettime() ordinals

tuday2:
Still need to write tests

### DIFF
--- a/MBBSEmu.Tests/ExportedModules/Majorbbs/gettime_Tests.cs
+++ b/MBBSEmu.Tests/ExportedModules/Majorbbs/gettime_Tests.cs
@@ -1,0 +1,42 @@
+﻿using MBBSEmu.HostProcess.Structs;
+using MBBSEmu.Memory;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+namespace MBBSEmu.Tests.ExportedModules.Majorbbs
+{
+    public class gettime_Tests : ExportedModuleTestBase
+    {
+        private const int GETTIME_ORDINAL = 327;
+
+        [Theory]
+        [InlineData(1, 1, 2000, 23, 0, 1)]
+        [InlineData(2, 2, 2010, 1, 30, 30)]
+        [InlineData(3, 3, 2020, 8, 45, 15)]
+        [InlineData(4, 22, 2035, 11, 0, 59)]
+        [InlineData(12, 31, 1980, 20, 59, 1)]
+        [InlineData(1, 18, 2038, 2, 14, 7)]
+        [InlineData(1, 1, 1970, 0, 0, 0)]
+        public void gettime_Test(int month, int day, int year, int hour, int minute, int second)
+        {
+            //Reset State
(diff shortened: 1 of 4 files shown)


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -7868,5 +7874,65 @@ private void txtvars_delegate()
...
+                Registers.AX = 0xFFFF; // -1
+                return;
+            }
+
+            var fileStructPointer = Module.Memory.GetOrAllocateVariablePointer($"FILE_{fileStreamPointer.Name}-{FilePointerDictionary.Count}", FileStruct.Size);
+            var fileStruct = new FileStruct(Module.Memory.GetArray(fileStructPointer, FileStruct.Size));
+
+            switch (fileMode)
+            {
+                case 0x4000: //TEXT
```
paladine, line 7905: in theory a client could send a bitmask with both TEXT & BINARY set, but it's probably such a fringe case. Maybe we could throw an exception?

MBBSEmu.Tests/ExportedModules/Majorbbs/gettime_Tests.cs
```diff
@@ -0,0 +1,42 @@
...
+            var dosTimeDateStructPointer = mbbsEmuMemoryCore.AllocateVariable("DOSDATETIME", TimeStruct.Size);
+            
+            //Execute Test
+            ExecuteApiTest(HostProcess.ExportedModules.Majorbbs.Segment, GETTIME_ORDINAL, new List<FarPtr> { dosTimeDateStructPointer });
+
+            //Verify Results
+            var dosTimeDateStruct = mbbsEmuMemoryCore.GetArray(dosTimeDateStructPointer, TimeStruct.Size);
+            
+            Assert.Equal(minute, dosTimeDateStruct[0]);
+            Assert.Equal(hour, dosTimeDateStruct[1]);
```
paladine, line 38: probably want to assert on 0 == dosTimeDateStruct[2]

paladine: unless of could we want to implement this using milliseconds / 10

paladine (changes requested):

MBBSEmu.Tests/ExportedModules/Majorbbs/gettime_Tests.cs
tuday2: We'll go with 0 :)

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
tuday2: Wouldn't the default: in the switch catch that exception? 
```
                default:
                    throw new Exception($"Unknown setmode: {fileMode}");
```

paladine (approved):

PR #420, opened 2021-03-01
Replace stzcpy implementation with strncpy + null terminate.

paladine:
stzcpy was overflowing by overreading the src pointer.


paladine:
Also updated MemoryCore to use a non-static Logger

enusbaum (approved):

PR #425, opened 2021-03-05
Fix for WG rlogin extra chars

synacktic:
Simple fix for the Worldgroup rlogin. Not much else to say really.

### DIFF
--- a/MBBSEmu/Extensions/BlockingCollectionExtensions.cs
+++ b/MBBSEmu/Extensions/BlockingCollectionExtensions.cs
@@ -1,0 +1,15 @@
+﻿using System.Collections.Concurrent;
+
+namespace MBBSEmu.Extensions
+{
+    public static class BlockingCollectionExtensions
+    {
+	public static void Clear<T>(this BlockingCollection<T> blockingCollection)
+	{
+	    while (blockingCollection.Count > 0)
+	    {
+		blockingCollection.TryTake(out _);
+	    }
+	}
+    }
+}
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -545,14 +545,16 @@ namespace MBBSEmu.HostProcess
         }
 
         /// <summary>
-        ///     Invokes routine registered as LONROU during MAJORBBS->REGISTER_MODULE() call on only
-        ///     the selected Rlogin module, specified by ModuleIdentifier.
+        ///     Invokes routine registered as LONROU during MAJORBBS->REGISTER_MODULE() call for
+        ///     all modules, and then enters module specified by ModuleIdentifier.
         ///
         ///     Executes on the given channel once after a user successfully logs in.
         /// </summary>
         private void ProcessLONROU_FromRlogin(SessionBase session)
         {
             session.OutputEnabled = false; // always disabled for RLogin
+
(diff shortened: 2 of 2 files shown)


enusbaum:
I don't think this extra data is part of the rlogin payload. I'm debugging at that point and here's the state during calling lonrou after connecting via rlogin from wg3nt:

image

enusbaum:
I can confirm the data comes in later after `lonrou` has been called.

Modifying `ProcessIncomingClientData` in `RloginSession.cs` with this code seems to work:

```csharp

 if (SessionState != EnumSessionState.Negotiating)
            {
                //Ugly WG3NT RLOGIN Extra Data Hack
                if (bytesReceived == 12 && clientData[0] == 255 && clientData[1] == 255)
                    return (null, 0);

                return (clientData, bytesReceived);
            }

```

It identifies a signature of the incoming weird data from Worldgroup's Rlogin client and drops it. 

A little hack-y, but it works? 🤷‍♂️

enusbaum (changes requested):
See comment on main PR thread

synacktic:
It was working on my side, so I'm guessing it just had to do with processor speed in that case. I'll look over the dirty hack, and use it or something like it. Would you like me to remove or leave the .clear() extension @enusbaum ?

synacktic:
Ok, now I think we wanted to add a config item, maybe "Rlogin.Behavior" or "Rlogin.Compat" or "Rlogin.Source" or some such to specify certain behaviors based upon the client that will be connecting.

One concern is, might someone want to connect from more than one type of system and need to be able to select certain behaviors?

I know which bit is a fix for WGNT, but I don't know what code might be doing work around for mystic and synchronet...

Mystic:  #76  ?
WG #425 
Synchronet ??

TLDR; What do we want this config setting to look like and what code will it enable / disable?

enusbaum:
Let's call it `Rloing.Compatibility` and we'll define an enumerator for `WG3NT`, and `None` (or `Default`)

We'll use the proposed hack but wrap it to only invoke if the `WG3NT` option is selected for compatibility. If the option is `None` or missing entirely, the compatibility code won't be invoked.

synacktic:
I'm sure I went about this in a funny way, but you can always course correct me.

MBBSEmu/Configuration.cs
```diff
@@ -64,6 +64,17 @@ public AppSettings(ILogger logger)
...
         public string RloginIPAddress => GetIPAppSettings("Rlogin.IP");
         
+        public Session.Rlogin.EnumRloginCompatibility RloginCompatibility
+        {
+            get
+            {
+                if (!ConfigurationRoot.GetSection("Rlogin.Compatibility").Exists())
+                    return Session.Rlogin.EnumRloginCompatibility.Default;
+                    
+                return (Session.Rlogin.EnumRloginCompatibility)Enum.Parse(typeof(Session.Rlogin.EnumRloginCompatibility), ConfigurationRoot["Rlogin.Compatibility"]);
```
enusbaum, line 74: Change this to an `Enum.TryParse` to be safer if they type in a value that can't be matched

enusbaum (changes requested):
Just the one change, looks good!

MBBSEmu/Configuration.cs
synacktic: It is not a drop in replacement, so I changed things around a bit.

enusbaum (approved):

PR #437, opened 2021-03-17
Add an Int21h Interrupt test and implement some more DOS interrupts

paladine:


tuday2 (approved):
Passed tests w/37 modules, code looks good

PR #439, opened 2021-03-18
Adds 32 bit support to ESI/EDI/ESP/EBP/EF

paladine:
Also some other minor fixes

- Moves EXE runtime to 0xA0000 address top
- Adds pushfd/popfd instruction
- Adds movsb instruction
- Inits RealModeMemoryCore to have all halt instructions in case CS:IP goes haywire


tuday2 (approved):
Passed 37 module test and let run for 1 hour (12 RTick's running) looks good!

enusbaum (approved):

PR #440, opened 2021-03-20
Bunch o'changes

paladine:
Added STC/LOOPNE
Added tests for STC/CLC/LOOP/LOOPE/LOOPNE
Fix bug in REP/REPE/REPNE - updated tests accordingly
Changed HLT to set Registers.Halt - modified RealModeMemoryCore to initialize everything with halts, so if something screws up, the system is halted
Added support for file handles in Int21h - can open pre-existing files now
Configured environment variable segment in EXEs
Fixed starting stack segment/pointer in EXEs
Shifted top memory address of EXEs to A000:0000 (video memory). It turns out extra space is needed for data segments and libc refused to initialize without adequate space after the end of the EXE memory
New int21 - 0x3D, 0x3E, 0x43,

### DIFF
--- a/MBBSEmu/DOS/ExeRuntime.cs
+++ b/MBBSEmu/DOS/ExeRuntime.cs
@@ -22,7 +22,7 @@ namespace MBBSEmu.DOS
         private ILogger _logger;
         private FarPtr _programRealModeLoadAddress;
 
-        private const int PROGRAM_MAXIMUM_ADDRESS = (0xF000 << 4);
+        private const int PROGRAM_MAXIMUM_ADDRESS = (0xA000 << 4);
         private const ushort PSP_SEGMENT = 0x2000;
 
         /// <summary>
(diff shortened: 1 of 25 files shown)


tuday2 (approved):
Passed 37 module test and let run for 1 hour (12 RTick's running) looks good!

MBBSEmu/DOS/ExeRuntime.cs
```diff
@@ -125,15 +137,15 @@ private void SetupPSP()
...
         {
-            if (_environmentVariables == null)
-                return;
-
             ushort bytesWritten = 0;
             foreach (var v in _environmentVariables)
             {
-                Memory.SetArray(ENVIRONMENT_SEGMENT, bytesWritten, Encoding.ASCII.GetBytes(v));
-                bytesWritten += (ushort)(v.Length + 1);
+                string str = v.Key + "=" + v.Value + "\0";
```
enusbaum, line 143: var

enusbaum (approved):

PR #443, opened 2021-03-21
Add EXE Name to Environment Segment after Variables

enusbaum:
Clears an Abnormal Abort on Borland TC++ Library Function which loads the EXE name at runtime

image

### DIFF
--- a/MBBSEmu/DOS/ExeRuntime.cs
+++ b/MBBSEmu/DOS/ExeRuntime.cs
@@ -145,7 +145,10 @@ namespace MBBSEmu.DOS
                 bytesWritten += (ushort)(str.Length);
             }
             // null terminate
-            Memory.SetWord(ENVIRONMENT_SEGMENT, bytesWritten, 0);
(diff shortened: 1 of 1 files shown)


MBBSEmu/DOS/ExeRuntime.cs
```diff
@@ -145,7 +145,10 @@ private void SetupEnvironmentVariables()
                 bytesWritten += (ushort)(str.Length);
             }
             // null terminate
-            Memory.SetWord(ENVIRONMENT_SEGMENT, bytesWritten, 0);
+            Memory.SetByte(ENVIRONMENT_SEGMENT, bytesWritten++, 0);
+
```
paladine, line 149: didn't that document you linked to me have a DWORD value of 0x0001 sitting between the end of environment and the EXE string?

paladine: Is that the 2 `inc si` calls?

PR #445, opened 2021-03-21
Tapatch3

paladine:
- Add cmpsb instruction, used for memcmp
- Overhauled shift instructions to properly set CF/OF (required for memcpy). OF is only set on shifts for 1 bit shifts
- Added int21 seek/read/write & general refactor (easier to set carry flag return status for example)


MBBSEmu.Tests/CPU/SHR_Tests.cs
```diff
@@ -0,0 +1,63 @@
+﻿using Iced.Intel;
+using MBBSEmu.CPU;
+using MBBSEmu.Extensions;
+using Xunit;
+using static Iced.Intel.AssemblerRegisters;
+
+namespace MBBSEmu.Tests.CPU
+{
+    public class SHR_Tests : CpuTestBase
```
enusbaum, line 9: We should add tests for shifts of up to the max value per type to ensure it keeps the correct byte with the type extension code. 

Also, this logic:

```
To reduce the maximum execution time, the 386 DX microprocessor does not allow shift counts greater 
than 31. If a shift count greater than 31 is attempted, only the bottom five bits of the shift count are 
used. (The 8086 uses all eight bits of the shift count.) 
```

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -166,6 +172,112 @@ public void Handle()
             }
         }
 
+        private void ClearCarryFlag() => _registers.F = _registers.F.ClearFlag((ushort)EnumFlags.CF);
```
enusbaum, line 175: Add the compiler optimization for `Subroutine` optimization so this gets inlined to avoid the additional call.

enusbaum (changes requested):

enusbaum (approved):

PR #448, opened 2021-03-24
Get EXEs fully loaded and initialized!

paladine:
- Adds `DAA` and `AAM` instructions to support `printf("%p")`
- Fixes XCHG bug related to xchg between reg & memory address
- Fixed PSP to be 256 bytes before main program
- Moved ENV_SEGMENT to be 4k before PSP


MBBSEmu/CPU/CPUCore.cs
```diff
@@ -1157,6 +1163,44 @@ private void WriteToSource(ushort result)
             }
         }
 
+        [MethodImpl(OpcodeCompilerOptimizations)]
+        private void Op_Aam()
+        {
+            var al = Registers.AL;
+            var imm = _currentInstruction.Immediate8;
+
+            Registers.AH = (byte)(al / imm);
```
enusbaum, line 1172: Should we use `imm`? Intel Docs say it's always divided by `10`

image

enusbaum (changes requested):

enusbaum (approved):

PR #452, opened 2021-03-26
INT 21h & INT 10h Enhancements for Console I/O

enusbaum:
- Refactored EXERuntime & INT 21h to use `LocalConsoleSession` (allows for easy porting of EXE's to output over Telnet, Rlogin, etc.)
- Added INT 21h APIs for Console I/O
- Added INT 10h APIs for Video Mode Information


MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -29,8 +30,8 @@ public class Int21h : IInterruptHandler
         public CpuRegisters Registers { get; set; }
         private IMemoryCore _memory { get; init; }
         private IClock _clock { get; init; }
-        private TextReader _stdin { get; init; }
-        private TextWriter _stdout { get; init; }
+        private BlockingCollection<byte> _stdin { get; init; }
```
paladine, line 33: Can we make this a simple interface rather than this specific type? I get the feeling this will be harder to use for testing purposes. I'm planning on creating a test that loads `CMDLINE.EXE` and controls the input/output via a `MemoryStream`

Doesn't have to be a heavyweight `TextReader`, but maybe something simpler like
```
public interface Reader {
  byte Read();
}

public interface Writer {
  void Write(byte c);
  void Write(byte[] c);
}
```
You could create a simple implementation on top of LocalSession. Thoughts?

PR #457, opened 2021-04-02
Restructure CPU registers in core to be a struct with overlapping memory areas. Avoids having to mask|or all operations on 16-bit register access. Speeds up benchmarks just shy of 10%

paladine:
Structs are CpuRegistersStruct
Interface is ICpuRegisters
CpuRegisters implements ICpuRegisters on top of CpuRegistersStruct
CpuCore inherits from CpuRegisters. So the core uses the internal
structs for fast updates, and interrupt handlers and other code use
ICpuRegisters to interact with the structs


enusbaum (approved):
LGTM reviewing, need to pull it and give it a run through

PR #463, opened 2021-04-05
Use LocalConsoleSession for Interrupt handling

paladine:
Let's do this

Also reworks memory layout to match DOS with executables


MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -201,6 +208,15 @@ private void SetCarryFlagErrorCodeInAX(DOSErrorCode code)
             Registers.AX = (ushort)code;
         }
 
+        private void DirectStdinInputNoEcho_0x07()
+        {
+            Registers.DL = 0xFF;
+            var c = _stdin.Read();
+            _stdout.Write(c);
+            Registers.AL = c;
+            Registers.F = Registers.F.ClearFlag((ushort)EnumFlags.ZF);
```
paladine, line 217: looks like this needs updating

MBBSEmu/DOS/Interrupts/Int21h.cs
```diff
@@ -201,6 +208,15 @@ private void SetCarryFlagErrorCodeInAX(DOSErrorCode code)
             Registers.AX = (ushort)code;
         }
 
+        private void DirectStdinInputNoEcho_0x07()
+        {
+            Registers.DL = 0xFF;
+            var c = _stdin.Read();
+            _stdout.Write(c);
```
paladine, line 215: no echo shouldn't write to stdout I suppose

enusbaum (approved):

PR #477, opened 2021-04-14
Run Cleanup/BBS UP commands as part of startup/cleanup

paladine:


MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -882,6 +893,29 @@ private void ProcessIncomingCharacter(SessionBase session)
             }
         }
 
+        private void RunProgram(string modulePath, string cmdline)
+        {
+            var exe = cmdline.Split(' ')[0];
+            var args = cmdline.Split(' ').Skip(1).ToArray();
+            exe = Path.ChangeExtension(exe, ".EXE");
+
+            exe = _fileUtility.FindFile(modulePath, exe);
```
enusbaum, line 902: This isn't persisting the path so when I run it locally in my IDE, it's looking in my IDE's current application directory, not the directory of the module. We probably need to pass in `modulePath` to the `MZFile` and persist that down the chain on the opens

This is the error I get when it's trying to establish the `MZFile`

image

enusbaum (changes requested):
Crashes on Windows/When the module isn't in the same directory as MBBSEmu

MBBSEmu/HostProcess/MbbsHost.cs
paladine: Nice catch - forgot to send the full path to MZFile

enusbaum (approved):

PR #481, opened 2021-05-01
Fix for MajorMUD 1.11p Help File (MSG)

enusbaum:
- Fixes mid-identification of data type of it's on its own line prior to the following identifier

MajorMUD 1.11p (WCCMMUD) uses a MSG file for parsing the help commands (using `sameto()`) and the help content. This file isn't exactly formatted consistently, specifically the MSG type definition is on its own line in several parts of the file:

image

vs

image

Because of this, our MSG parsing code was identifying the `String` string as the following identifier, and since it contained `spaces` and then alpha characters for the ACTUAL following identifier, it set the state to `JUNK` for the actual identifier when it was encountered. This caused the offset of the MSG ordinals to be off and all kinds of bad memory overflows to happen.

This SHOULD fix it, but it's not resilient as a MSG identifier followed by a `\r` before `{` will be thrown away as junk. This probably needs some testing.

### DIFF
--- a/MBBSEmu/Module/MsgFile.cs
+++ b/MBBSEmu/Module/MsgFile.cs
@@ -108,7 +108,7 @@ namespace MBBSEmu.Module
                     case MsgParseState.IDENTIFIER when IsIdentifier(c):
                         identifier.Append(c);
                         break;
-                    case MsgParseState.IDENTIFIER when c == '\r':
+                    case MsgParseState.IDENTIFIER when c is '\r' or '\n':
                         state = MsgParseState.JUNK;
                         break;
                     case MsgParseState.IDENTIFIER when char.IsWhiteSpace(c):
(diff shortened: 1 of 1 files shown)


MBBSEmu/Module/MsgFile.cs
```diff
@@ -108,6 +108,9 @@ private void BuildMCV()
                     case MsgParseState.IDENTIFIER when IsIdentifier(c):
                         identifier.Append(c);
                         break;
+                    case MsgParseState.IDENTIFIER when c == '\r':
```
paladine, line 111: `when c == '\r' || c == '\n'` would be better here I think, in case the MSG is formatted in Unix format without `\r`

Though I don't think this parser would succeed without `\r` anyhow, there are other places that look suspect

paladine (changes requested):

PR #489, opened 2021-05-30
MBBSEmu FSE Implementation

enusbaum:
Initial MajorBBS/Worldgroup Full Screen Editor (FSE) Implementation. We're re-using the Full Screen Display (FSD) code by using a custom template & logic for FSE. 

This is a very basic implementation. We're not handling things like word wrap, max length checks, etc. yet -- but this first pass should get `bgnedt()` working.


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            var topicLength = GetParameter(3);
+            var topicPointer = GetParameterPointer(4);
+            var onDoneEditing = GetParameterPointer(6);
+            var flags = GetParameter(8);
+
+            //This is where we'll build the Answers String to pass into the FSD Template
+            var answersString = new StringBuilder(topicLength + textLength);
+
+            //Compile a Topic Answer String
+            answersString.Append($"TOPIC={Encoding.ASCII.GetString(Module.Memory.GetString(topicPointer))}");
```
paladine, line 8113: can use the `StringBuilder(string? value, int capacity)` constructor in lieu of `new + Append`

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            
+            //Take the input text and split it out on new lines to we can hydrate the individual lines on the template
```
paladine, line 8115: to -> so?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            var rawText = Encoding.ASCII.GetString(Module.Memory.GetString(textPointer, true));
+            var split = rawText.Split('\r');
+            var lineNumber = 0;
+            for (var i = 0; i < split.Length; i++)
+            {
+                if (split[i].Length < 79)
+                {
+                    answersString.Append($"LINE{i}={split[i]}\0");
```
paladine, line 8123: should be `" LINE"` (with space) like below, or no space in the string here?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            //Double Null Terminated
+            answersString.Append('\0');
+
+            //FSDROOM
+            var resourceManager = new ResourceManager();
+
+            var template = resourceManager.GetResource("MBBSEmu.Assets.fseTemplate.ans");
+            var fieldSpec = resourceManager.GetResource("MBBSEmu.Assets.fseFieldSpec.txt");
+
+            var fsdBufferPointer = Module.Memory.GetOrAllocateVariablePointer($"FSD-TemplateBuffer-{ChannelNumber}", 0x2000);
```
paladine, line 8146: `0x2000` = 8k which is pretty darn huge no?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            var fsdFieldSpecPointer = Module.Memory.GetOrAllocateVariablePointer($"FSD-FieldSpec-{ChannelNumber}", 0x2000);
+
+            //Zero out FSD Memory Areas
+            Module.Memory.SetZero(fsdBufferPointer, 0x2000);
+            Module.Memory.SetZero(fsdFieldSpecPointer, 0x2000);
+
+            //Hydrate FSD Memory Areas with Values
```
paladine, line 8153: Hydrate - you keep using that word, I do not think it means what you think it means

https://i.kym-cdn.com/entries/icons/original/000/010/692/19789999.jpg

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            Module.Memory.SetArray(fsdBufferPointer, template);
```
paladine, line 8154: + null terminators?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+                fldspc = fsdFieldSpecPointer,
+                flddat = fsdfldPointer,
+                newans = newansPointer
+            };
+
+            Module.Memory.SetArray($"FSD-Fsdscb-{ChannelNumber}", fsdStatus.Data);
+
+            //FSDAPR
+            var fsdAnswersPointer = Module.Memory.GetOrAllocateVariablePointer($"FSD-Answers-{ChannelNumber}", 0x800);
+            Module.Memory.SetArray($"FSD-Answers-{ChannelNumber}", Encoding.ASCII.GetBytes(answersString.ToString()));
```
paladine, line 8175: + null terminator?

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            var fsdscbStruct = new FsdscbStruct(Module.Memory.GetArray(fsdscbStructPointer, FsdscbStruct.Size));
+            
+            Module.Memory.SetArray(fsdscbStructPointer, fsdscbStruct.Data);
+
+            //Set Exit Routine
+            var fsdWhenDoneRoutine = Module.Memory.GetOrAllocateVariablePointer($"FSD-WhenDoneRoutine-{ChannelNumber}", FarPtr.Size);
+            Module.Memory.SetPointer(fsdWhenDoneRoutine, onDoneEditing);
+
+            //Set Pointers for Topic & Text
+            Module.Memory.GetOrAllocateVariablePointer($"FSE-TextPointer-{ChannelNumber}", FarPtr.Size);
```
paladine, line 8195: nit: chain results so you don't repeat the strings

`Module.Memory.SetPointer(Module.Memory.GetOrAllocateVariablePointer(...), value);`

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8084,5 +8091,117 @@ private void gettime()
...
+            Module.Memory.SetPointer($"FSE-TextPointer-{ChannelNumber}", textPointer);
+            Module.Memory.GetOrAllocateVariablePointer($"FSE-TopicPointer-{ChannelNumber}", FarPtr.Size);
+            Module.Memory.SetPointer($"FSE-TopicPointer-{ChannelNumber}", topicPointer);
+
+#if DEBUG
+            _logger.Debug($"Channel {ChannelNumber} entering Full Screen Editor");
+#endif
+
+            Registers.Halt = true;
```
paladine, line 8204: maybe a comment here would be useful for others, like `prevents module from continuing until we're done`

MBBSEmu/HostProcess/HostRoutines/FsdRoutines.cs
```diff
@@ -348,35 +390,35 @@ private void EnteringFullScreenDisplay(SessionBase session)
...
         /// </summary>
-        /// <param name="userInput"></param>
+        /// <param name="session"></param>
         private ushort SetUserInput(SessionBase session)
         {
             var userInput = session.InputBuffer.ToArray();
 
             if (userInput.Length == 0 && session.CharacterProcessed == 0x8)
-                userInput = new byte[] {0x8};
+                userInput = new byte[] { 0x8 };
```
paladine, line 399: comment here would be good `0x8 == backspace`

MBBSEmu/HostProcess/HostRoutines/FsdRoutines.cs
```diff
@@ -596,7 +637,27 @@ private void InFullScreenDisplay(SessionBase session)
...
+                                //Final Field in FSE is for Savings
+                                if (session.SessionState == EnumSessionState.InFullScreenEditor)
+                                {
+                                    var fseFinalState =
+                                        session.CurrentModule.Memory.GetOrAllocateVariablePointer(
+                                            $"FSE-FinalStateCode-{session.Channel}", sizeof(ushort));
+                                    switch (_fsdFields[session.Channel].SelectedField.Value)
+                                    {
+                                        case "SAVE":
+                                            session.CurrentModule.Memory.SetWord(fseFinalState, 0);
```
paladine, line 650: can fseFinalState be null?

MBBSEmu/HostProcess/HostRoutines/FsdRoutines.cs
```diff
@@ -627,19 +687,57 @@ private void ExitingFullScreenDisplay(SessionBase session)
             session.CurrentModule.Memory.SetZero(fsdscbStruct.newans, 0x800);
             session.CurrentModule.Memory.SetArray(fsdscbStruct.newans, Encoding.ASCII.GetBytes(answersString));
```
paladine, line 688: + null terminate?

paladine (changes requested):

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: No spaces in the string. The Answers are null terminated strings and the answer string itself it double null terminated.

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: Yes, but my math there was (80*24) just for the raw ASCII characters possible (1920), then potentially ANSI sequences consuming more of the space as well (up to 6-7 additional bytes per character, depending). I don think an FSD template would use an ANSI sequence per character -- but possible. This just accounts for the edge case.

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: 💧

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: We set zero prior to the areas declares and then fill them in. So while these Span's aren't null terminated, the area they're writing to has been null populated.

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: `answerString` is compiled as double null terminated prior

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
enusbaum: Personally I try not to chain many functions into a single line as it hampers debugging where I'd have to `step into` several potential functions to debug my nested one.

MBBSEmu/HostProcess/HostRoutines/FsdRoutines.cs
enusbaum: The pointer? No, it's declared prior to the switch.

The logic here is basically saying, "if you're on the last field of the FSE FSD template AND hit enter, evaluate the input" as it's always the QUIT/SAVE multi-select. The valid return values according to the docs are `0` and `-1`

MBBSEmu/HostProcess/HostRoutines/FsdRoutines.cs
enusbaum: The pattern I'm using for writing Spans to these arrays is setting zero to the destination array prior to writing a ASCII result. It's probably lazy -- but guarantees it'll be null terminated 😛

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
paladine: line 8131 has a space still

paladine (approved):

PR #492, opened 2021-06-03
Only Trigger "Status Changed" if Status == 240 or 5

### LINKED ISSUE #490: [LUNATIX] Lunatix - Stack at initial splash screen, can't progress into game
**Module Information**
* LUNATIX
* Lunatix
* 5.3f

**Describe the bug**
Can't get past intro screen after launching game. Feels like a regression as this used to work in https://github.com/mbbsemu/MBBSEmu/releases/tag/v1.0-alpha-010421

**To Reproduce**
1. Launch MBBSEmu
2. Either telnet/rlogin into the MBBSEmu menu _or_ directly to the rlogin game port for Lunatix
3. See ansi splash screen/info screen
4. Press enter or any other key to continue and note that you're stuck at this screen

**Expected behavior**
Game logic continues and allows you to either create a character (if you haven't already) or continue gaming from previous login

**Screenshots**
image

**Software Information:**
- OS: Ubuntu 20.04 amd64
- Telnet Client: NetRunner as well as linux telnet and linux rlogin

enusbaum:
Lunatix would call `btuinj` on entry and set the status to `240` (deferred execution). This would trigger the `StatusChanged` property. Because of this, input wouldn't be processed. I've added code to only trigger the `StatusChanged` property if it's set to deferred execution, otherwise handle it as regular input.

There's potentially a larger refactor that needs to happen here, where channel `Status` is handled as a FIFO queue, because Lunatix is calling `btuinj` to set `Status` of `240`, then setting status of `1` on routine "exit". This means the first actual status read from the channel SHOULD be `240`, which would trigger `stsrou`. 

The full refactor would be setting `STATUS` by dequeuing the most recently available status from the Status queue in the `MbbsHost` loop, and persisting that for the lifetime of the loop. 

For now -- I think this will work.

Fixes #490

### DIFF
--- a/MBBSEmu/HostProcess/ExportedModules/Galgsbl.cs
+++ b/MBBSEmu/HostProcess/ExportedModules/Galgsbl.cs
@@ -321,9 +321,9 @@ namespace MBBSEmu.HostProcess.ExportedModules
 
             //Status Change
             //Set the Memory Value
-            Module.Memory.SetWord(Module.Memory.GetVariablePointer("STATUS"), status);
+            Module.Memory.SetWord("STATUS", status);
 
-            //Notify the Session that a Status Change has occured
+            //Notify the Session that a Status Change has occurred
             channel.StatusChange = true;
 
 #if DEBUG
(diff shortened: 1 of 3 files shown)


enusbaum:
More Info:

image

PR #495, opened 2021-07-05
Status fifo queue

### LINKED ISSUE #490: [LUNATIX] Lunatix - Stack at initial splash screen, can't progress into game

enusbaum:
Change Status from a `ushort` to a `Queue<ushort>`. MajorBBS modules can inject status codes using `btuinj` while still also manually setting `STATUS` to `1` (ok) after `sttrout` runs, which would have previously overwritten the injected status. Moving to a queue means we no longer need to track if a status has "changed", we just process what's in the queue.

Fixes #490


MBBSEmu/HostProcess/HostRoutines/MenuRoutines.cs
```diff
@@ -225,8 +224,7 @@ private void LoginPasswordDisplay(SessionBase session)
         private void LoginPasswordInput(SessionBase session)
         {
             //Only Process on CR
-            if (session.Status != 3) return;
-            session.Status = 0;
+            if (session.GetStatus() != 3) return;
```
paladine, line 227: we really need to use some enums here to get rid of these magic number statuses

paladine (approved):

PR #502, opened 2021-10-04
MBBSEmu SoftPatch Support

enusbaum:
This PR adds support to MBBSEmu to "SoftPatch" modules, where modifications to modules can be loaded via config without having to apply changes directly to DLL files.

- Added `Patch` option to the Module Configuration JSON Specification
- Supported patch types are `Hex` (string of Hex Codes to be patched in) and `Text` (string to be patched in directly)
- Changed Module JSON Loading to straight JSON Deserialization using `System.Text.Json` (versus loading it as a configuration file via `System.Configuration`)
- Added Unit Tests for new Module Configuration Loading + New JSON Converters
- Updated `modules.json` that is included to reflect how to apply a SoftPatch

Currently only `AbsoluteOffset` is supported, I'll add support for `SEGMENT:OFFSET` patches in another PR, but the properties are defined here.

This will allow for easy byte level patching of binaries without having to use a hex editor.

bitmoji


MBBSEmu/Converters/JsonBooleanConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+                case JsonTokenType.False:
+                    return false;
+                default:
+                    throw new ArgumentOutOfRangeException();
+            }
+        }
+
+        public override void Write(Utf8JsonWriter writer, bool value, JsonSerializerOptions options)
+        {
+            switch (value)
```
paladine, line 54: 9 lines down to 1?
`return writer.WriteStringValue(value ? "1" : "0");`

paladine (approved):

paladine (approved):

PR #510, opened 2021-10-31
Fix Cracked! game

tuday2:
-Stubbed out Ordinal 82
-Stubbed out Ordinal 549
-Fixed secure echo bug in signup
-Fixed secure echo bug in bad password
-Added output messaging/logging for a denial of access related to a key


tuday2:
Fixes #510

enusbaum (approved):

PR #514, opened 2021-11-29
Dotnet6

enusbaum:
Submitting PR to test build


enusbaum:
OK -- We're Green. Will look into the Node deprecation warnings later, but looks like we're building!

paladine:
Checks out on Linux, approved

paladine (approved):

PR #517, opened 2021-12-23
MajorMUD 1.11p for DOS Fixes (WCCMMUD)

enusbaum:
With the refactor to handle `STATUS` as a FIFO queue, in `ProcessIncomingCharacter()`, when `0xD` was detected, it wasn't enqueuing a status of `3` unless the queue was empty. Because MajorMUD uses polling, the `STATUS` queue contained `192`, meaning the input ready status of `3` wasn't being added.

I changed the logic to *always* enqueue status of `3` on `0xD` input, after it does the BTUCHI input check. This is the correct action.

Fixing this re-enables input on MajorMUD

I also added a couple more friendly error messages/help when the MajorBBS/Worldgroup module isn't configured or installed properly.


MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -857,22 +857,18 @@ private void ProcessIncomingCharacter(SessionBase session)
...
                         if (session.GetStatus() == 240)
                         {
                             //Set Status == 3, which means there is a Command Ready
                             session.Status.Clear(); //Clear the 240
                             session.Status.Enqueue(3); //Enqueue Status of 3
                             session.EchoSecureEnabled = false;
+                            break;
                         }
 
+                        //Always Enqueue Input Ready
```
paladine, line 870: would changing the default to 3 do the same thing?

https://github.com/mbbsemu/MBBSEmu/blob/3257d2cc8390077d8e8e69949b3e43d6b99b95de/MBBSEmu/Session/SessionBase.cs#L339

enusbaum: No -- if we defaulted the status to 3, it'd assume there was input to process on each iteration of the event loop and trigger `STTROU`. We default it to 1 (OK/READY) which lets us know there's nothing that needs to be done.

paladine (approved):

PR #518, opened 2021-12-24
Btrieve File Size Validation/Sanity Check

enusbaum:
Checks for minimum length of the Btrieve file so the header check doesn't crash if the file is 0 bytes (for whatever reason)


paladine (approved):
I was going to say let's add some unit test coverage, but we don't appear to support that. BtrieveFile only supports files and not random byte arrays, so maybe I can rework it a bit for better unit testing

PR #519, opened 2021-12-24
Enable VT100 Terminal on Windows Console when running external EXE

enusbaum:
We need to force VT100 mode when running an external EXE (even if not using Local Console Session), so the ANSI output from the EXE being executed is properly handled by the local console and not just written out as a ton of ANSI code text.

Submitting this PR to strike an experimental build -- waiting for reporter of the issue to respond if this fixes it.

### DIFF
--- a/MBBSEmu/DOS/Win32VT100.cs
+++ b/MBBSEmu/DOS/Win32VT100.cs
@@ -1,7 +1,7 @@
 ﻿using System;
 using System.Runtime.InteropServices;
 
-namespace MBBSEmu.Session.LocalConsole
+namespace MBBSEmu.DOS
 {
     /// <summary>
     ///     Enables VT100 Terminal Emulation with Command Console by calling SetConsoleMode through an external call to the Kernel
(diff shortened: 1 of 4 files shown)


MBBSEmu/DOS/Win32VT100.cs
```diff
@@ -0,0 +1,74 @@
...
+
+                if (!SetConsoleMode(iStdIn, inConsoleMode))
+                    throw new Exception($"Failed to set Input Console Mode, Error Code: {GetLastError()}");
+
+                if (!SetConsoleMode(iStdOut, outConsoleMode))
+                    throw new Exception($"Failed to set Output Console Mode, Error Code: {GetLastError()}");
+            }
+            catch (Exception e)
+            {
+                _logger.Error(e);
```
paladine, line 69: Can be combined right?

`_logger.Error("Blah blah blah", e);`

paladine (approved):

MBBSEmu/DOS/Win32VT100.cs
enusbaum: The only reason I ordered it like this is so the "friendly" error message doesn't get lost at the top of the stack trace/exception dump, and the friendly message is below it on the console -- so would be the first message the person sees when scrolling back up. More a QOL thing

PR #522, opened 2021-12-31
Change user status to an enum, better than magic numbers

paladine:
Branches from linebreaks, so merge that one first


MBBSEmu/HostProcess/Enums/UserStatus.cs
```diff
@@ -0,0 +1,37 @@
+namespace MBBSEmu.HostProcess.Enums
+{
+    /// <summary>
+    /// User Status values, mostly taken from BRKTHU.H
+    /// </summary>
+    public enum UserStatus
```
enusbaum, line 6: Rename to `EnumUserStatus` to keep consistent Enum naming convention

MBBSEmu/HostProcess/Enums/UserStatus.cs
```diff
@@ -0,0 +1,37 @@
...
+    {
+        UNUSED = 0, // default unused value
```
enusbaum, line 8: You can use XML documentation on the enums to provide tool tips for their value while in an IDE

MBBSEmu/HostProcess/Enums/UserStatus.cs
enusbaum, line 6: I also don't _THINK_ it matters, but Enum's are typed as `int` by default, but we can specify ` : ushort` just to keep the type consistent.

enusbaum (changes requested):

MBBSEmu/HostProcess/Enums/UserStatus.cs
paladine: I thought Polish notation was dead?

enusbaum (approved):

Issue #523, opened 2022-01-01
[WCCMMUD] Majormud - Can't queue multiple commands


fletcherm:
**Module Information**
* WCCMMUD
* Majormud
* 1.11p mbbsemu-friendly version

**Describe the bug**

A Majormud can't queue multiple commands that should execute one after another.

For example, if I quickly type "w enter w enter w enter" I should move west three times.

In mbbsemu, doing the above will add all three "w" commands together, and try to execute "www", which is not a valid command. So Majormud interprets it as a say command.

Similarly, if you were to type "i enter a rat enter", it should be interpreted as the inventory command (i) followed by a command to attack a rat (a rat), but will all come out as one say command.

**To Reproduce**

Try to move several rooms in rapid succession.

**Expected behavior**

Commands should be queued to execute after the prior command finishes.

**Screenshots**

![unknown](https://user-images.githubusercontent.com/147532/147854798-b1cdbb8e-696a-4d94-b316-659d4f9b29e3.png)

enusbaum:
I believe the issue here is that MegaMUD is sending `w\nw\n`, but MBBSEmu ignores `\n` on Input:

https://github.com/mbbsemu/MBBSEmu/blob/master/MBBSEmu/HostProcess/MbbsHost.cs#L876

Telnet clients send `\r` for `ENTER`, so in this case, it's filtering out the `\n` and handling the input as `ww\r`

We could change the input processor to change instances of `\n` to `\r` when processing input for `STTROU`, and then only ignore if the PREVIOUS input was `\r` (to prevent double input on `\r\n`)

fletcherm:
Looks good on the 2022-01-02 build. Thanks!

PR #525, opened 2022-01-02
Input fixes for MMUD

paladine:
Push polling status & then pop it during poll routine

Ignore TransparentMode when handling \r on input

Fix line break crash related to backspace/deletes

Also fixed INPLEN to not account for extra null character - it was always +1 too large


enusbaum:
Fixes #523

enusbaum (approved):

PR #529, opened 2022-01-03
Make word breaks/wrapper contingent on btutsw

paladine:


MBBSEmu/HostProcess/ExportedModules/Galgsbl.cs
```diff
@@ -684,12 +684,20 @@ private void btuech()
         /// </summary>
         private void btutsw()
         {
-            var channel = GetParameter(0);
+            var channelNumber = GetParameter(0);
```
enusbaum, line 687: We can remove the XML comment for the method that this is _ignored_ 😃

enusbaum (changes requested):
BASICALLY UNUSABLE 😆

PR #533, opened 2022-01-05
Local config sanity

synacktic:
Currently changing our local configs for testing can be a little annoying because they are in git too. Let's rename them to *.sample, so we don't have to worry about that anymore.


synacktic:
Lemme do that again

PR #534, opened 2022-01-05
Local config sanity

synacktic:
Currently changing our local configs for testing can be a little annoying because they are in git too. Let's rename them to *.sample, so we don't have to worry about that anymore.


synacktic:
@enusbaum

enusbaum:
I think we should add LaunchSettings to GitIgnore.

If we're going to rename the `appconfig.json`, we might want to add some verbiage to the CLI default help screen about the file containing some default settings that they can rename.

A-OK with renaming modules.json to sample.

synacktic:
Added the launchsettings.json to .gitignore, if we can see a way to rename appsettings.json.sample for packaging, then we should be golden.

enusbaum:
Just adding this so we don't lose track:

This PR needs to also enhance the build YAML to perform a rename on the `appsettings.json.sample` to `appsettings.json` so we have a file available to run by default for folks who download the release packages.

enusbaum:
Closing -- this might cause more confusion.

PR #537, opened 2022-01-09
Soft patch enhancement

enusbaum:
- Add FarPtr Converter for JSON
- Add `Address` and `Addresses` to specify SEG:OFF for patch locations
- Added constructor to FarPtr that takes `SEG:OFF` string
- Added CRC32 calculation on Module and verification of CRC32 on Patches (to ensure correct version is being patched)
- Prints CRC32 of Module Loaded on Startup
- Added `Enabled` as an option on Patches so they can be easily enabled/disabled without having to delete them from the file

### DIFF
--- a/MBBSEmu/Module/ModulePatch.cs
+++ b/MBBSEmu/Module/ModulePatch.cs
@@ -21,6 +21,7 @@ namespace MBBSEmu.Module
         public uint AbsoluteOffset { get; set; }
         public EnumModulePatchType PatchType { get; set; }
         public string Patch { get; set; }
+        public string CRC32 { get; set; }
 
         /// <summary>
         ///     Returns bytes for the patch depending on the Patch Type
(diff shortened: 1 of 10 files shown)


enusbaum:
OK -- ready for real this time 😉

MBBSEmu/Module/ModulePatch.cs
```diff
@@ -15,11 +16,18 @@ public enum EnumModulePatchType
 
         public string Name { get; set; }
         public string Description { get; set; }
-        public ushort Segment { get; set; }
-        public ushort Offset { get; set; }
+        public List<FarPtr> Addresses { get; set; }
+        public FarPtr Address { get; set; }
```
paladine, line 20: seems a bit silly/error-prone to support a single Address as well as a List of them. How about just a List, and if you want one item, it's just a list of one item?

MBBSEmu/Util/CRC32.cs
```diff
@@ -0,0 +1,119 @@
+﻿// Copyright (c) Damien Guard.  All rights reserved.
```
paladine, line 1: why is this not built into the .NET framework already, yikes

paladine (approved):

PR #538, opened 2022-01-10
Wbtrv32.dll implementation in C++/CLI

paladine:


enusbaum (approved):
We might also want to add a README to this directory with details on this project

PR #557, opened 2022-03-30
Fix MSG parse issue with CPTDTXT.MSG

paladine:
MSG parsing code would enter a junk state when encountering a \r\n
immediately after an identifier, but it should allow the transition
to the SPACE state instead. Technically \r\n are space characters
after all.


enusbaum:
This would undo #481

We might need to add some logic to evaluate if an identifier has been extracted (is the `\r\n` before or after) as the two cases seem distinct.

enusbaum:
Also I'm wondering if we should abstract out the parser to have a similar `Tick()` command (`Step()`?) for parsing, as it would allow us to pass in potentially a string/file and Unit Test state given a specific scenario.

enusbaum:
Re-Wrote the MSG File Loader/Parser. It's much more flexible now and also supports Unit/Integration testing. I tested this PR on `TSGARN`, `WCCMMUD`, and `CPTDAWN` and it loads all the MSG file values properly.

MBBSEmu/Module/MsgFile.cs
```diff
@@ -40,13 +40,37 @@ public MsgFile(IFileUtility fileUtility, string modulePath, string msgName)
...
+        {
+            /// <summary>
+            ///     Data Before Identifier Characters
+            /// </summary>
+            PREKEY,
+
+            /// <summary>
+            ///     Data for Identifier
+            /// </summary>
+            KEY,
```
paladine, line 53: comment has identifier, so why not `IDENTIFIER?`

PR #559, opened 2022-04-27
Additional logging on Socket

enusbaum:
Added Additional Logging on the Socket to record any additional exceptions that might happen on the socket for debugging

### DIFF
--- a/MBBSEmu/Session/SocketSession.cs
+++ b/MBBSEmu/Session/SocketSession.cs
@@ -57,6 +57,11 @@ namespace MBBSEmu.Session
             {
                 _logger.Warn($"Channel {Channel}: Attempted to write on a disposed socket");
             }
+            catch (Exception ex)
+            {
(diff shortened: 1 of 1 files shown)


MBBSEmu/Session/SocketSession.cs
```diff
@@ -57,6 +57,11 @@ public virtual void Send(byte[] dataToSend)
             {
                 _logger.Warn($"Channel {Channel}: Attempted to write on a disposed socket");
             }
+            catch (Exception ex)
+            {
+                _logger.Error($"Channel {Channel}: Error Sending Data");
+                _logger.Error(ex);
```
paladine, line 63: should we also rethrow in these cases? Otherwise the exception is just swallowed and might not invoke future catch handlers

enusbaum:
An unhandled exception in the `Send()` and `Receive()` threads will crash the main process. At this point it's more for additional debugging _if_ there's an issue there.

PR #576, opened 2023-05-05
Sscanf enhancements

paladine:
Adds support for % lengths and also %c support


enusbaum (approved):
Tested locally and verified unit tests, looks good

PR #579, opened 2023-07-09
Update RNG to use Borland C++ RNG Routine

enusbaum:
- Update `srand()` to set seed in memory
- Change `rand()` from using internal C# RNG to ported Borland C++ routine


enusbaum:
Verified against Borland C++ 4.5

Compiled the following using `bcc rand.cpp`:

```c++
#include <stdlib.h>
#include <stdio.h>
#include <time.h>

int main() {
    // Seed the random number generator
    srand(1234);

    // Generate and output the next three random numbers
    for (int i = 0; i < 10; ++i) {
        int randomNum = rand();
        printf("Random number %d: %d\n", (i + 1), randomNum);
    }

    return 0;
}
```

Output Results:

image

Unit Tests for `rand()` are verified against this list

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -8334,5 +8339,20 @@ private ushort setINPLEN(FarPtr input)
...
         /// </summary>
         private ReadOnlySpan<byte> hdlcon => Module.Memory.GetVariablePointer("HDLCON").Data;
+
+        /// <summary>
+        ///     Sets the Seed for the Pseudo-Random Number Generator
+        ///
+        ///     While the seed stored in memory is a 32-bit long, only the lower 16-bits are set
+        ///     via the srand() method. The high 16-bits are always set to zero.
+        ///
+        ///     Signature: void srabd(int seed)
```
paladine, line 8349: Signature: void `srand`(int seed)

paladine (approved):

PR #585, opened 2023-08-21
Print warning messages before cleanup runs.

fletcherm:
Note: this code was written largely by ChatGPT 4, so it may look a little different than the surrounding code. It was, of course, still inspected by a human. :)

What appears to be working:
* A warning message is sent 5 minutes before cleanup.

What appears to be missing:
* Even though the code is setup to send more messages every minute, we're only seeing that initial 5 minute warning message. This needs further investigation.

Next steps:
* Ensure a warning message is sent every minute before cleanup.
* (Possibly) adding some tests for this. This file doesn't already have a test suite so not clear to me if the juice is worth the squeeze.
* Compare the color of the cleanup warning text to MBBS proper. Right now we're using the same red color as the final cleanup message, but not sure if this is correct.
* Compare the literal text of the final cleanup message. My spidey sense is that does not match MBBS proper.

### DIFF
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -155,8 +155,12 @@ namespace MBBSEmu.HostProcess
             _exportedFunctions = new Dictionary<string, IExportedModule>();
             _realTimeStopwatch = Stopwatch.StartNew();
             _incomingSessions = new Queue<SessionBase>();
+
+            //Setup Cleanup Restart Event
             _cleanupTime = _configuration.CleanupTime;
-            _cleanupTimer = new Timer(_ => _performCleanup = true, this, NowUntil(_cleanupTime), TimeSpan.FromDays(1));
+            _cleanupTimer = new Timer(_ => _performCleanup = true, null, NowUntil(_cleanupTime), TimeSpan.FromDays(1));
+            Logger.Info($"Waiting {NowUntil(_cleanupTime)} until {_cleanupTime} to perform Nightly Cleanup");
+
             _cleanupWarningTimer = SetupCleanupWarningTimer();
 
             if (_configuration.TimerHertz > 0)
@@ -1398,15 +1402,19 @@ namespace MBBSEmu.HostProcess
             Start(moduleConfigurations);
         }
 
+        /// <summary>
+        ///     Returns a TimeSpan representing the time between now and the specified time.
+        /// </summary>
+        /// <param name="timeOfDay">24-hour timespan representing a time of day, and determining the time between now and the time specified</param>
+        /// <returns></returns>
         private TimeSpan NowUntil(TimeSpan timeOfDay)
         {
-            var waitTime = _cleanupTime - Clock.Now.TimeOfDay;
+            var waitTime = timeOfDay - Clock.Now.TimeOfDay;
             if (waitTime < TimeSpan.Zero)
             {
                 waitTime += TimeSpan.FromDays(1);
             }
 
-            Logger.Info($"Waiting {waitTime} until {timeOfDay} to perform nightly cleanup");
             return waitTime;
         }
 
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -228,7 +228,7 @@ namespace MBBSEmu.HostProcess
         public void Stop()
         {
             _isRunning = false;
-            _cleanupTimer.Dispose();
+            _cleanupTimer?.Dispose();
             _cleanupWarningTimer?.Dispose();
             _tickTimer?.Dispose();
             // this set must come after _isRunning is set to false, to trigger the exit of the
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -104,7 +104,7 @@ namespace MBBSEmu.HostProcess
         private Timer _cleanupWarningTimer;
 
         /// <summary>
-        ///     Amount of time to start warning before cleanup.
+        ///     Amount of time to start warning before cleanup grace period ends.
         /// </summary>
         private const int CleanupWarningInitialMinutes = 5;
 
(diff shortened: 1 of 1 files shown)


fletcherm:
@enusbaum @paladine This is a draft PR and not ready to merge. But I'd appreciate if you can take a look and at least see if you can identify when I'm only seeing the first warning message and not subsequent messages.

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -212,7 +228,8 @@ public void Start(List<ModuleConfiguration> moduleConfigurations)
         public void Stop()
         {
             _isRunning = false;
-            _timer.Dispose();
+            _cleanupTimer.Dispose();
+            _cleanupWarningTimer?.Dispose();
```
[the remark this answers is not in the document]
enusbaum: I think we should add the null check to both. While they're assigned in the constructor used most everywhere, if we decide to overload the constructor in the future (Unit Testing, etc.) we'll want to make sure this is gracefully handled.

Good catch!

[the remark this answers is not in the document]
fletcherm: Should be all set now.

fletcherm:
cleanup-warnings

Attached is a screenshot of what the cleanup warnings look like on WG.

fletcherm:
@enusbaum @paladine Alright, would love if you guys can take another look at this.

Only significant change since last round of reviews is that I changed how the cleanup time and grace period work, as we discussed in discord. (grace period comes _after_ cleanup time, for whatever reason) This should better match WG.

My bigger ask is that someone test & confirm the behavior that cleanup and the cleanup warning run more than once, and not just after the initial setup. Normally that would be once a day, of course, but feel free to tweak values to make it happen more frequently.

I _did_ do some experiments to test this myself, and things look good... but I had to hack enough lines of code to make cleanup more frequent that I kinda worry I changed something else incidentally and tainted the experiment. That is why I'd really appreciate another person testing it as well.

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -1287,16 +1313,63 @@ private void DisableModule(string moduleId)
...
 
+        private Timer SetupCleanupWarningTimer()
+        {
+            _cleanupWarningMinutesRemaining = CleanupWarningInitialMinutes;
+
+            var initialWarningTime = _cleanupTime + _cleanupGracePeriod - TimeSpan.FromMinutes(CleanupWarningInitialMinutes);
+            var dueTime = NowUntil(initialWarningTime);
+            var period = TimeSpan.FromMinutes(1);
+
+            Logger.Debug($"_cleanupTime Hours: {_cleanupTime.Hours}");
```
enusbaum, line 1324: We can probably remove these. While they're helpful developing this feature, I don't see us needing these printed every startup in debug.

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -1287,16 +1313,63 @@ private void DisableModule(string moduleId)
...
+            {
+                Logger.Debug($"in SendCleanupWarning, _cleanupWarningMinutesRemaining = {_cleanupWarningMinutesRemaining}");
+
+                var minuteText = (_cleanupWarningMinutesRemaining > 1) ? "minutes" : "minute";
+
+                foreach (var c in _channelDictionary)
+                {
+                    var channel = c.Value.Channel;
+
+                    Logger.Error($"Sending cleanup warning to channel {channel}: {_cleanupWarningMinutesRemaining} {minuteText} until shutdown.");
```
enusbaum, line 1346: This can be changed to DEBUG or INFO

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -1287,16 +1313,63 @@ private void DisableModule(string moduleId)
...
+                    _channelDictionary[channel].SendToClient($"|RESET|\r\n|B||MAGENTA|Sorry to interrupt here, but the server will be shutting down in {_cleanupWarningMinutesRemaining} {minuteText} for the nightly \"auto-cleanup\" process. Please finish up and log off... thank you!|RESET|\r\n".EncodeToANSIArray());
+                }
+                _cleanupWarningMinutesRemaining--;
+                Logger.Debug($"finished SendCleanupWarning");
```
enusbaum, line 1350: Can be removed

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -96,14 +96,36 @@ public class MbbsHost : IMbbsHost, IDisposable
...
+
+        /// <summary>
+        ///     Track minutes left until the nightly cleanup occurs.
+        /// </summary>
+        private int _cleanupWarningMinutesRemaining = CleanupWarningInitialMinutes;
+
+        /// <summary>
+        ///     Amount of time to give as a grace period to logoff after nightly cleanup.
+        /// </summary>
+        private const int _cleanupGracePeriodMinutes = 10;
```
enusbaum, line 119: Following our naming for constants, this should be `CleanupGracePeriodMinutes`

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -1287,16 +1313,63 @@ private void DisableModule(string moduleId)
...
+
+        private void SendCleanupWarning(object _)
+        {
+            if (_cleanupWarningMinutesRemaining > 0)
+            {
+                Logger.Debug($"in SendCleanupWarning, _cleanupWarningMinutesRemaining = {_cleanupWarningMinutesRemaining}");
+
+                var minuteText = (_cleanupWarningMinutesRemaining > 1) ? "minutes" : "minute";
+
+                foreach (var c in _channelDictionary)
```
enusbaum, line 1342: You can combine the channel selection on the next line and the `foreach` definition into one line using a LINQ expression for the values being iterated over: 

`foreach (var channel in _channelDictionary.Select(c => c.Value.Channel))`

enusbaum (changes requested):

enusbaum (approved):

PR #586, opened 2023-09-01
New Setup UI

enusbaum:
Adds a new cross platform, text based GUI used to configure initial settings in the `appsettings.json` file.

Establishes a UI Base and repeatable process for adding console GUI's to MBBSEmu in the future.

Also updated NuGet Packages for the project

image
image
image

### DIFF
--- a/MBBSEmu/AppSettingsManager.cs
+++ b/MBBSEmu/AppSettingsManager.cs
@@ -1,0 +1,343 @@
+using Microsoft.Extensions.Configuration;
+using NLog;
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.IO;
+using System.Linq;
+using System.Net;
+using System.Text.Json;
+
+namespace MBBSEmu
+{
+    /// <summary>
+    ///     Handles Application Settings and Configuration for MBBSEmu
+    ///
+    ///     Configuration Values are loaded from the specified appsettings.json file (command line) or the default
+    /// </summary>
+    public class AppSettingsManager
+    {
+        public readonly IConfigurationRoot ConfigurationRoot;
+
+        private readonly ILogger _logger;
+
+        /// <summary>
+        ///     Safe loading of appsettings.json for Configuration Builder
+        /// </summary>
+        /// <returns></returns>
+        public AppSettingsManager(ILogger logger)
+        {
+            _logger = logger;
+
+            if (!File.Exists(ConfigurationFileName))
+                throw new FileNotFoundException($"Unable to locate [{ConfigurationFileName}] emulator settings file.");
+
+            if (!IsValidJson(File.ReadAllText(ConfigurationFileName)))
+                throw new InvalidDataException($"Invalid JSON detected in [{ConfigurationFileName}]. Please verify the format & contents of the file are valid JSON.");
+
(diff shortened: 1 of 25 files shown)


MBBSEmu/AppSettingsManager.cs
```diff
@@ -0,0 +1,343 @@
...
+                .Build();
+        }
+
+        /// <summary>
+        ///     Configuration File being used by MBBSEmu
+        /// </summary>
+        public string ConfigurationFileName => Program._settingsFileName ?? Program.DefaultEmuSettingsFilename;
+
+        //Validate Config File
+        public string BBSTitle => GetBBSTitleSettings("BBS.Title");
```
paladine, line 49: we setup all those mappings in `AppSettings.cs` but they're not reused here at all which leads to duplication and makes me sad

MBBSEmu/UI/Setup/SetupView.cs
```diff
@@ -0,0 +1,517 @@
...
+
+            wizard.Finished += (args) =>
+            {
+                //args.Cancel = true;
+            };
+
+            wizard.Cancelled += (args) =>
+            {
+                //args.Cancel = true;
+            };
```
paladine, line 62: rm?

paladine (approved):

MBBSEmu/AppSettingsManager.cs
enusbaum: There might not be a way around this one. We handle the config file once loaded through the dotnet Configuration handler via `Microsoft.Extensions.Configuration`, but needed the new class to serialize/validate the JSON generated by the new Wizard. We could probably refactor the use of the configuration runtime out in a future PR, since we abstract it away anyway.

MBBSEmu/UI/Setup/SetupView.cs
```diff
@@ -0,0 +1,517 @@
...
+                HelpText = "Welcome to The MajorBBS Emulation Project!\n\n" +
+                                 "MBBSEmu was created to allow quick, easy access to the Games we all love and remember that were written for The Major BBS 6.25 and Worldgroup for DOS!\n\n" +
+                                 "Our project is Open Source, Free, and Community Supported!\n\n" +
+                                 "This Setup Wizard will guide you through the initial setup of MBBSEmu and the base settings. Click or Select Next below to start!",
+                ColorScheme = _wizardStepColorScheme
+            };
+            wizard.AddStep(firstStep);
+
+            //Next Step - General Settings
+            var secondStep = new Wizard.WizardStep("General Settings")
```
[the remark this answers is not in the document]
enusbaum: I lifted the setup routine and style from the Terminal.GUI example project ([https://github.com/gui-cs/Terminal.Gui/blob/develop/UICatalog/Scenarios/WizardAsView.cs](https://github.com/gui-cs/Terminal.Gui/blob/develop/UICatalog/Scenarios/WizardAsView.cs)), and while I could break it out into its own method -- I'd need an intermediary class (like a List) to hold the step objects before adding them to the Wizard in order to setup the validation events at the end.

I don't feel this is a show stopper, more just a readability thing as I don't feel we'll be reusing any steps in the Wizard elsewhere.

PR #588, opened 2023-09-17
New Console GUI and Logger Implementation

enusbaum:
Adding a new "main" UI when running MBBSEmu

This is a QOL enhancement using Termina.Gui, allowing for a cross-platform terminal UI. We'll be able to add functionality into this UI, such as:

- Ability to Add/Remove Modules while running
- Gracefully recover/debug modules that crash while running without having to restart
- Enable/Disable services through the UI
- Build out a "Modules Configuration Builder" which simplifies the task of creating a `modules.json` file for multiple modules
- Online CPU Debugging/Stepping through OpCodes + Register Values

A major refactor in this version is removing `NLog` and using our own custom logger implementation which is pretty lightweight and allows us to easily handle logging for different scenarios. Custom loggers for Audit Trail, CPU Debugging, etc.

image
image
image


MBBSEmu/Logging/CPULogger.cs
```diff
@@ -0,0 +1,38 @@
...
+namespace MBBSEmu.Logging
+{
+    /// <summary>
+    ///     Logger used for logging CPU Instructions, Registers, and Messages
+    ///
+    ///     Messages are in turn written to the specified Logging Targets
+    /// </summary>
+    public class CpuLogger
+    {
+        private static readonly List<ILoggingTarget> LOGGING_TARGETS = new();
```
paladine, line 15: ew, global static state

MBBSEmu/Logging/Targets/ConsoleTarget.cs
```diff
@@ -0,0 +1,26 @@
...
+        public void Write(params object[] logEntry)
+        {
+            if(logEntry.Length != 0)
+                return;
+
+            var logMessage = (string)logEntry[0];
+            var logLevel = ((EnumLogLevel)logEntry[1]).ToString();
+
+            //Use reflection to get the name of the class calling this method
+            var callingType = new System.Diagnostics.StackTrace().GetFrame(3)?.GetMethod()?.DeclaringType;
```
paladine, line 21: at least in Java, retrieving the stack trace is an expensive operation. If lots of logging occurs it'll bog down performance I'd imagine

MBBSEmu/Logging/Targets/IQueueTarget.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace MBBSEmu.Logging.Targets
+{
+    public interface IQueueTarget
+    {
+        object[] Dequeue();
+        IList<object[]> DequeueAll();
+        void Clear();
+    }
```
paladine, line 10: make Generic?

MBBSEmu/Program.cs
```diff
@@ -280,28 +294,29 @@ private void Run(string[] args)
...
+                    //Start UI in a Task as to not block this thread
+                    Task.Run(() =>
+                    {
+                        mainMBBSEmuWindow.Setup();
+                        mainMBBSEmuWindow.Run();
+                    });
+
+                    //Wait for the UI to be running before continuing
+                    while (!mainMBBSEmuWindow.isRunning)
+                        Thread.Sleep(100);
```
paladine, line 311: maybe sleep longer and add some sort of console out to let people know something is happening, like {
  Console.print("Waiting for main window to create");
  Thread.Sleep(1000);
}

paladine (approved):

MBBSEmu/Logging/Targets/ConsoleTarget.cs
enusbaum: image

MBBSEmu/Logging/CPULogger.cs
enusbaum: image

MBBSEmu/Logging/Targets/ConsoleTarget.cs
enusbaum: I think in a refactor we can move the reflection of the calling type up the stack into the Logger. I imagine this is why you have to pass in `GetLogger<T>` in most logging libraries in order to save it from having to look up the calling type. Our release version doesn't log nearly as much where I think this will be of any material performance impact, but I'll mark it with a `TODO` so we don't lose track of it.

MBBSEmu/Logging/Targets/IQueueTarget.cs
enusbaum: C# doesn't support unknown generics like Java (we discussed in Discord), and technically this accomplished the same thing since `object` will need to be boxed to the destination type. My goal was to keep the Targets generic enough as to where a given target can be assigned to any Logger type.

MBBSEmu/Program.cs
enusbaum: Made it sleep 500ms and it appends a `.` to the console message `Waiting for MBBSEmu GUI to start...`

I wouldn't expect this to append more than 1-2 period before the GUI thread fully starts

PR #596, opened 2023-11-05
Fix RCL Implementation

enusbaum:
- Fixes RCL Implementation to be accurate
- Added Unit tests for both 8-bit and 16-bit RCL operations
- Minor code cleanup

### DIFF
--- a/MBBSEmu/CPU/CPUCore.cs
+++ b/MBBSEmu/CPU/CPUCore.cs
@@ -349,7 +349,7 @@ namespace MBBSEmu.CPU
             };
 
             //Set this value to TRUE if you want the CPU to break after each instruction
-            var CPUDebugBreak = true;
+            var CPUDebugBreak = false;
 
             //Evaluate Breakpoints
             if (CPUBreakpoints.Contains(_currentInstructionPointer))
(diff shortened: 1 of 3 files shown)


MBBSEmu/CPU/CPUCore.cs
```diff
@@ -1665,17 +1669,17 @@ private byte Op_Rcl_8()
...
 
-                    //Perform Rotation
+                    // Perform Rotation
                     result = (byte)(result << 1);
 
-                    //If CF was set, rotate that value in
+                    // If CF was set, rotate that value in to the LSB (0th bit)
                     if (Registers.CarryFlag)
-                        result.SetFlag(1);
+                        result |= 1;
```
paladine, line 1680: isn't this equivalent?

paladine (approved):

MBBSEmu/CPU/CPUCore.cs
enusbaum: No, technically `.SetFlag` is an extension method we have on `byte` types, which since they're static, can't modify the underlying instance value. This was one of the bugs with the underlying logic here, because if using the helper method it should be `result = result.SetFlag(1)`, which I felt was overhead for just a ` |= 1` operation.

paladine: ah, subtle

PR #597, opened 2023-11-06
Modules Config "BasePath"

enusbaum:
The goal with this PR is that we can specify one `BasePath` in a Module Configuration File. In the near future, this will make packaging a "MBBSEmu Game Pack" easier to distribute as every module can be relatively pathed off a root folder with a single configuration. 

Additionally, it should make managing a Modules Config JSON file with multiple modules easier to manage if all the modules share the same relative path.

- Add `BasePath` Option to Module Configuration JSON
- Create new JSON ModuleConfiguration Converter that loads Modules from the JSON and applies Base Path to each deserialized Module
- ModuleConfiguration will now apply `BasePath` to a modules specified path if it is set, and the Module path specified isnt a  Root path


enusbaum:
Oh -- I guess I could write some Unit Tests for this as well  :)

paladine (approved):
sure whatever

PR #609, opened 2023-12-08
Debug & Crash Report Fixes

enusbaum:
- Fixes an Unhandled Exception in Crash Report (inconsistent user of `.ToHexString()` extension)
- Added additional error handling to `.ToHexString()` extension
- Added Unit Tests for `ReadOnlySpan<>()` extension class
- Add Register Values to CPU Debug Output

### DIFF
--- a/MBBSEmu.Tests/Extensions/ReadOnlySpanExtensions_Tests.cs
+++ b/MBBSEmu.Tests/Extensions/ReadOnlySpanExtensions_Tests.cs
@@ -1,0 +1,62 @@
+﻿using Xunit;
+using MBBSEmu.Extensions;
+using System;
+
+namespace MBBSEmu.Tests.Extensions
+{
+    /// <summary>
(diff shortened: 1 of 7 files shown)


MBBSEmu.Tests/Extensions/ReadOnlySpanExtensions_Tests.cs
```diff
@@ -0,0 +1,62 @@
...
+            Assert.False(result);
+        }
+
+        [Fact]
+        public void ToHexStringTest()
+        {
+            ReadOnlySpan<byte> input = new byte[] { 1, 2, 3, 4, 5 };
+            var result = input.ToHexString(0, 5);
+            Assert.DoesNotContain("No Data to Display", result);
+            Assert.DoesNotContain("Invalid Address Range", result);
```
paladine, line 43: can we assert on what it should be, rather than what it is not?

paladine (changes requested):

paladine (approved):

PR #615, opened 2023-12-31
Apply Default Keys on -DBRESET

enusbaum:
The Command Line Argument `-DBRESET` was applying a set of default keys on reset, but not the actual default keys defined in `appsettings.json`. 

This change will apply a fixed set of keys to Sysop (`SUPER`, `SYSOP`) on reset, and then also apply the default keys defined in `appsettings.json`. The `GUEST` account will only have the default keys.

### DIFF
--- a/MBBSEmu/Database/Repositories/RepositoryBase.cs
+++ b/MBBSEmu/Database/Repositories/RepositoryBase.cs
@@ -14,11 +14,9 @@ namespace MBBSEmu.Database.Repositories
     ///
     ///     Holds common Dapper Functionality for executing Queries
     /// </summary>
-    public abstract class RepositoryBase : IRepositoryBase, IDisposable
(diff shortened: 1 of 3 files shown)


MBBSEmu/Database/Repositories/RepositoryBase.cs
```diff
@@ -14,32 +14,24 @@ namespace MBBSEmu.Database.Repositories
     ///
     ///     Holds common Dapper Functionality for executing Queries
     /// </summary>
-    public abstract class RepositoryBase : IRepositoryBase, IDisposable
+    public abstract class RepositoryBase(ISessionBuilder sessionBuilder, IResourceManager resourceManager) : IRepositoryBase, IDisposable
     {
```
paladine, line 18: is this some fancy C# way to define fields in the constructor rather than inside the class itself?

paladine (approved):

PR #620, opened 2024-01-14
Correct the name of the mbbsemu.db file; add message at the end reminding the user to recreate the BBSUSR database.

fletcherm:
Correct the name of the mbbsemu.db file; add message at the end reminding the user to recreate the BBSUSR database.


paladine (changes requested):
The name of the config file is specified in appsettings.json as `Database.File`

Please make this a configurable flag in `_create_parser` and you can set the default to mbbsemu.db

paladine (approved):

PR #627, opened 2024-07-01
Don't log the user in if we're AT max, not _ABOVE_

enusbaum:
- Do not allow a user to connect via Rlogin if we're _AT OR ABOVE_ max defined channels, not just _ABOVE_


tuday2 (approved):
Nice fix! This will correctly limit the # of users based on the value set for # of BBS channels without crashing.

PR #629, opened 2024-07-24
Variable Ordinals

enusbaum:
- Fix for functions that return strings (or memory pointers) to return unique values for each returned value within a single execution loop
- Ordinals Reset to 0 on each execution loop, allowing for memory to be reused


enusbaum:
Verified that there are only two methods missing unit tests: `scanmsg` and `rawmsg`

Upon further investigation, the current `scanmsg` implementation is incorrect. We only support reading from a MSG file loaded (and defined) in the current module MDF, but the documentation states it should be _ANY_ MSG file. 

So we'll need to correct that, I'll open a defect.

paladine (approved):

PR #653, opened 2026-02-27
Harden SQLite runtime stability in the Btrieve layer

mparrett:
> **Parked pending #662.** #662 rewrites `BtrieveFileProcessor.cs`, which this branch changes by 200 lines, so the two conflict directly. Once #662 lands this will be reverified against the new storage layer — much of it is likely to be moot when Btrieve is no longer SQLite-backed.
>
> **The non-Btrieve fixes moved to #679** — the `lonrou` cross-module fix, both `strcpy` fixes, the dictionary-mutation fix, the `FsdUtility` null check, and the `TELL` stub. That PR is conflict-free against #662 and reviewable now. The branch here is unchanged, so those commits still appear in the diff below until #679 merges.

## Why

MajorMUD's world state is write-heavy — 16,768 writes to `WCCMP001.DB` in one session against 14 to `WCCUSERS.DB` — and under that load a continuously running instance hit `SQLITE_BUSY` failures that killed the emulator. Four `SQLite Error 5: 'database is locked'` crashes between 2025-12-31 and 2026-01-09, all thrown from `BtrieveFileProcessor.Delete()` by way of `delbtv()`.

## What changed

**`SQLITE_BUSY` crashed Btrieve writes.** The connection now opens with `journal_mode = WAL` and `busy_timeout = 5000`, and commits retry on `SQLITE_BUSY` with a 50/200/500 ms backoff before giving up and rolling back.

The host runs a single worker thread across all channels, so a commit that exhausts every retry can stall the whole loop for roughly 20 seconds. I chose that over the crash, since a stall is recoverable; a shorter `busy_timeout` is the better call if you would rather fail fast.

**Rollback could throw inside the catch.** A failed `COMMIT` followed by a throwing `Rollback()` crashed while handling the original error, losing it. `SafeRollback` swallows the secondary failure so the first survives to be logged.

**`ReadOnlySpan<T>` compared against null.** `fieldSpec == null` and `key == null` are always false on a value type, so those guards never ran. Replaced with the intended length checks. The matching `FsdUtility` fix is on #679.

## Verification

Deferred until #662 merges, since the file this changes is rewritten there. For the record, the isolated SQLite commits pass the full suite on current `master`: 2599 passed, 0 failed, as of 2026-08-22. The Btrieve tests run noticeably slower under WAL mode with retries — roughly 8 minutes versus 3 — which is worth weighing against the crash it prevents.

### DIFF
--- a/MBBSEmu/Btrieve/BtrieveFileProcessor.cs
+++ b/MBBSEmu/Btrieve/BtrieveFileProcessor.cs
@@ -788,10 +788,10 @@ namespace MBBSEmu.Btrieve
                 currentQuery = new BtrieveQuery(this)
                 {
                     Key = Keys[(ushort)keyNumber],
-                    KeyData = key == null ? null : new byte[key.Length],
+                    KeyData = key.IsEmpty ? null : new byte[key.Length],
                 };
 
-                if (key != null)
+                if (!key.IsEmpty)
                 {
                     Array.Copy(key.ToArray(), 0, currentQuery.KeyData, 0, key.Length);
                 }
--- a/MBBSEmu/Btrieve/BtrieveFileProcessor.cs
+++ b/MBBSEmu/Btrieve/BtrieveFileProcessor.cs
@@ -276,23 +276,25 @@ namespace MBBSEmu.Btrieve
         /// </summary>
         private void UpgradeDatabaseFromVersion1To2()
         {
-            var transaction = Connection.BeginTransaction();
-            // creates the missing triggers
-            var keys = LoadSqliteKeys(transaction);
-            CreateSqliteTriggers(transaction, keys.Values);
-
-            // and bump the version
-            // not using GetSqliteCommand since this is used once and caching it provides no benefit
-            using var cmd = new SqliteCommand("UPDATE metadata_t SET version = 2", Connection, transaction);
-            cmd.ExecuteNonQuery();
+            using var transaction = Connection.BeginTransaction();
 
             try
             {
+                // creates the missing triggers
+                var keys = LoadSqliteKeys(transaction);
+                CreateSqliteTriggers(transaction, keys.Values);
+
+                // and bump the version
+                // not using GetSqliteCommand since this is used once and caching it provides no benefit
+                using var cmd = new SqliteCommand("UPDATE metadata_t SET version = 2", Connection, transaction);
+                cmd.ExecuteNonQuery();
+
                 transaction.Commit();
             }
-            catch (Exception)
+            catch (Exception ex)
             {
-                transaction.Rollback();
+                SafeRollback(transaction, "UpgradeDatabaseFromVersion1To2");
+                throw new InvalidOperationException("Failed to upgrade sqlite DB from version 1 to 2", ex);
             }
         }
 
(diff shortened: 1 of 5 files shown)


mparrett:
@enusbaum Will share more context about the scenario that inspired these changes. Thanks for your patience.

mparrett:
> @enusbaum Will share more context about the scenario that inspired these changes. Thanks for your patience.

Hi... sorry again for the long radio silence. Here's the context I owed on this one, from my last flurry of local tinkering— one Dockerized instance running MajorMUD (`WCCMMUD`) and Trade Wars 2002 (`HVSTW`) side by side, later T-LORD too, left up with players connected on telnet. 108 crash reports between 2025-12-30 and 2026-02-28. I can provide a more detailed account if you like, but it's already quite a large surface area for review. The fixes are that crash list, not a code audit:

- **`lonrou`** — `Module.DoLoginRoutine=False` suppressed output but still made the call, and Trade Wars' login routine breaks MajorMUD's command dispatch from there ("Your command had no effect"). Skipping the call fixes it. I did not isolate which state gets clobbered.
- **strcpy** — login crash: a destination segment that doesn't exist throws `IndexOutOfRangeException`, past the existing `ArgumentException` catch. Seen with a target in `0x1001`.
- **Dictionary mutation** — nightly cleanup can throw mid-enumeration and skip module disposal on the restart path.
- **SQLite** — MajorMUD's world state is write-heavy (16,768 writes to `WCCMP001.DB` in one session against 14 to `WCCUSERS.DB`); I hit `SQLITE_BUSY` plus rollbacks throwing from inside the catch block.
- **ReadOnlySpan** — the one item with no crash behind it, found while reading the Btrieve key path.

Correction to my note above, in case it reached you by email: I flagged an `Update()` error-mapping issue (commit failure surfacing as `InvalidKeyNumber`). That was stale — it came from an older review of this branch, and the SQLite commit in this PR already returns `IOError` on commit failure. There's nothing there to fix, so no follow-up PR.

On shape: I'd keep the current six together unless splitting reviews more easily for you.

paladine:
BtrieveFileProcessor is completely rewritten as part of https://github.com/mbbsemu/MBBSEmu/pull/662 - perhaps we should merge that first and then reverify this PR?

mparrett:
Good catch — this changes 200 lines of `BtrieveFileProcessor.cs`, so it conflicts with #662 directly.

Rather than parking all six behind it, I'm splitting. Four of them never touch Btrieve: the `lonrou` cross-module fix, the strcpy `IndexOutOfRangeException` catch, the dictionary-mutation fix, and the strcpy tests. Those move to a separate PR you can review independently of #662.

What's left is the SQLite hardening — WAL, busy retries, transaction failures. That waits for #662 and gets reverified against the new storage layer, and I'd expect much of it to be moot once Btrieve isn't SQLite-backed.

Working on the split now.

mparrett:
@paladine Split this out. The four non-Btrieve fixes moved to #679, and that what remains is the SQLite hardening likely to be moot once Btrieve isn't SQLite-backed.

MBBSEmu/Btrieve/BtrieveFileProcessor.cs
paladine, line 1: This should be reverted pending https://github.com/mbbsemu/MBBSEmu/pull/662?

paladine (changes requested):

mparrett:
Yes — and since this branch is now only the Btrieve changes, reverting them leaves nothing behind, so I'll close it rather than push an empty PR.

In case it didn't surface on your end: I split this earlier today. The four fixes that never touched Btrieve — `lonrou`, both `strcpy` fixes, and the dictionary-mutation fix, plus the `TELL` stub — moved to #679. That one is conflict-free against #662 and reviewable now. What stayed here was just the SQLite hardening.

I'll reopen this against the rewritten layer once #662 lands. I'd expect much of it to be moot by then, since the WAL and busy-retry work hardens a SQLite path #662 largely removes.

PR #654, opened 2026-03-01
Fix FSD display and field positioning

mparrett:
## Summary
- Fix StripAnsi to handle all ANSI sequence types (cursor, clear, etc.) not just color codes
- Fix FSD field extraction stopping at wrong character
- Fix CR counted as column width in field positioning
- Add cursor home after clear screen in fsdbkg()
- Skip echo for FullScreen states (FSD handles its own echo)
- Add bounds checking to prevent crash on field navigation past array bounds
- Fix FSD answer mapping crash when field count doesn't match expected indices

## Test plan
- [x] Unit tests pass (2501 tests)
- [ ] Test MajorMUD character creation (full-screen form)
- [ ] Test sysop bulletin editor
- [ ] Verify field navigation doesn't crash on edge cases

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/MBBSEmu.Tests/HostProcess/Fsd/FsdUtility_Tests.cs
+++ b/MBBSEmu.Tests/HostProcess/Fsd/FsdUtility_Tests.cs
@@ -1,0 +1,49 @@
+using MBBSEmu.HostProcess.Fsd;
+using System.Collections.Generic;
+using Xunit;
+
+namespace MBBSEmu.Tests.Fsd
+{
+    public class FsdUtility_Tests
+    {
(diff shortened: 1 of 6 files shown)


enusbaum (approved):
If this works in MajorMUD, I'll merge it in. That's the only major module we've used so far that makes extensive use of the FSD

mparrett:
@enusbaum Thanks! I have tested it quite a bit and can confirm it works. There are a couple of nits to fix in a future PR, but it's a net improvement so I support the merge.

PR #656, opened 2026-04-21
Memory Leaks fixes

seant-me:
I was able to fix resource leaks in session teardown and module cleanup. This should prevent the memory issues I experienced.

- Dispose CancellationTokenSource in SocketSession.CloseSocket() after Cancel() to release unmanaged resources on every session disconnect
- Dispose active file search enumerators (fnd1st/fndnxt) in Majorbbs.Dispose() to prevent file handle leaks when searches are abandoned before completion
- Use Dispose() instead of Close() on FileStream and Socket for idiomatic IDisposable cleanup


enusbaum (approved):
lgtm

PR #662, opened 2026-08-17
Wbtrv32 interop

paladine:
HAVE MBBSEMU USE WBTRV32.DLL

This depends on WBTRV32.DLL being built and in your path, or whether a WBTRV32_PATH environment variable is set to the path of the DLL. We'll probably have to come up with some smart way of integrating these, perhaps merging the projects? The automation pipeline will fail, but at least you can start the code review.


enusbaum (approved):
yolo

PR #664, opened 2026-08-18
Arm and consume GSBL screen-pause markers at the host

mparrett:
> **Rebased onto `master` — this no longer stacks on #652.** It previously carried #652's commits, which inflated the diff; it now contains only its own three commits and can be reviewed and merged independently, in any order relative to #652.

## Why

Modules that use GSBL's screen-pause feature embed control bytes the host is meant to consume and never transmit. T-LORD embeds Control-S (0x13, the clear-pause-counter character) throughout its screens and registers Control-T (0x14, the pause character) via `btupbc`. MBBSEmu never implemented the subsystem, so both bytes leaked to the client — raw 0x13/XOFF in native mode (which can freeze flow-control-honoring terminals), and `‼` glyphs once CP437→UTF-8 conversion renders them.

Byte-level comparison against real MajorBBS 6.25 wire captures shows the key behavior: **The Major BBS arms these markers host-side for every channel.** The real wire carries zero of them from the module's first output byte, on screens where T-LORD never calls `btupbc` — registration is an override, not the enabling event. (An earlier revision of this branch gated consumption on the module's `btupbc` call; an end-to-end test driving the full T-LORD journey showed that gate never engages while the embedded 0x13s leak, so it was replaced with host-armed defaults.)

image

## Change

- Sessions start with the Major BBS defaults armed: pause = 0x14, clear-pause-counter = 0x13.
- `Galgsbl.btupbc(chan, pausch)` overrides the pause character per channel (`pausch == 0` disables screen pause). `Galgsbl.btucpc(chan, clrpch)` (ordinal 86, previously undispatched) overrides the clear-pause-counter character independently. Exiting a module restores the defaults, so one door's override doesn't follow the user into the next.
- `SessionBase.SendToClient` consumes the armed markers from ASCII-mode output. Full-screen (FSD) output is GSBL binary mode, where the same byte values are CP437 display glyphs (e.g. MajorMUD's character-worksheet arrows): consumption is skipped during the full-screen session states and around `fsdbkg`'s template paint, which occurs while the session is still in-module.
- The blanket DC1–DC4 filter in `CreatePrintableCharacterArray` is removed — those bytes are now handled per-session by the mechanism above.

## Scope

Consuming the markers is the wire-correctness fix; the interactive pause prompt itself is not implemented (T-LORD's embedded 0x13s exist precisely to suppress it, and modern terminals have scrollback). Full `btuxmt`/`btuxmn` ASCII-vs-binary output-mode modeling remains future work; the full-screen session states plus an explicit binary-output flag cover every case observed.

## Verification

- `btupbc_Tests` cover the semantics: defaults armed at session start and consumed from ASCII output, 0x11/0x12 untouched, `btupbc`/`btucpc` overrides and 0-disable, binary-mode and full-screen passthrough, and defaults restored on module exit.
- New wire-invariant integration tests drive the real modules in-process, opt-in via a `MBBSEMU_TEST_MODULE_PATH` environment variable pointing at a directory of module installs (skipped otherwise, including CI, since module binaries are not redistributable): the full T-LORD journey — intro, character creation, stats, happenings, Town Square — transmits zero 0x11–0x14 bytes, matching the real-MBBS captures; MajorMUD's FSD character worksheet keeps its glyphs.
- Full suite green with the module tests enabled: 2580 passed, 0 failed.

The branch carries two small enabling commits: a byte-exact test harness (raw-byte session read APIs and golden-capture assertions — `GetLine`'s ASCII decoding destroyed every byte ≥ 0x7F, making CP437 wire assertions impossible) and an idempotent `BtrieveFileProcessor.Dispose` (a double-dispose NRE through the shared global-cache path, exposed by running real modules under test).


MBBSEmu.Tests/Util/ModuleFactAttribute.cs
paladine, line 1: Having tests that are conditional based on the installation is brittle. Usually we make the necessary changes to https://github.com/mbbsemu/MBBSEmu.Module to add whatever tests we need, and then write tests against those since we can include that module in our repo. Can we do that? I have a dev environment in a dos VM to compile whatever changes are necessary to MBBSEmu.Module

paladine (changes requested):

MBBSEmu.Tests/Util/GoldenAssert.cs
```diff
@@ -0,0 +1,128 @@
...
+    /// </summary>
+    public static class GoldenAssert
+    {
+        /// <summary>
+        ///     Asserts actual matches the golden file byte-for-byte, except within
+        ///     maskedRanges (offset/length spans covering volatile data such as
+        ///     dates, usernames, or random rolls), which are zeroed on both sides
+        ///     before comparison.
+        /// </summary>
+        public static void MatchesGolden(byte[] actual, string goldenPath, IEnumerable<(int Offset, int Length)> maskedRanges = null)
```
paladine, line 24: do we need to actually write data to the file and then turn around and immediately read it again? Seems like a waste that'll just slow down our tests with unnecessary I/O. Couldn't we just change the goldenData argument from a string to a byte array?

paladine (changes requested):

MBBSEmu.Tests/Util/GoldenAssert.cs
paladine: I'll just clean it up after the fact

paladine (approved):

PR #677, opened 2026-08-20
Fail CI when formatting does not match .editorconfig

mparrett:
Draft: the job fails on its own branch, for a reason worth resolving before this is worth merging. See "Blocked on SDK pinning" below.

Stacked on #676 and includes its commit, so merge that one first. Once it lands this PR reduces to the workflow change alone.

## Why

An `.editorconfig` only holds if something checks it. Without a gate, the tree drifts out of compliance one PR at a time and nobody notices until the next solution-wide cleanup, which is how #661 came about.

## What this does

Adds a `format` job to the existing workflow that runs `dotnet format whitespace --verify-no-changes`. It is independent of `build`, so a formatting failure reports on its own rather than masking a build or test result.

The scope is whitespace rather than a bare `dotnet format`, which would also run the analyzers. That distinction matters today: `CA2265` currently fires on `master`, so the unscoped command fails, and gating CI on it would force a decision about analyzer enforcement through a CI change instead of on its own merits.

## Blocked on SDK pinning

`dotnet format` does not produce the same output across SDK feature bands. This job resolved `10.0.x` to SDK 10.0.400 and failed on `CPUCore.cs`, asking for different indentation of the `goto` label than SDK 10.0.101 asks for locally. The branch is clean under one and dirty under the other.

So a formatting gate is only meaningful once the SDK is pinned, presumably through `global.json` so contributors, CI, and `docker/Dockerfile` all agree. Pinning `dotnet-version` in this workflow alone would make CI deterministic while leaving everyone's local runs free to disagree with it, which moves the problem rather than fixing it.

Adding `global.json` constrains every contributor's toolchain, so that seems like your decision rather than something to fold into a CI change. Happy to add it here, send it separately, or drop this PR if you would rather not gate on formatting at all.

## Verification

- `dotnet format whitespace --verify-no-changes` exits 0 on this branch and 2 on `master` under SDK 10.0.101. Under 10.0.400 the branch does not pass, which is the blocker above.
- The workflow parses, and the job matches the conventions already in the file: `actions/checkout@v6`, `actions/setup-dotnet@v5`, and `dotnet-version: 10.0.x`.

If you would rather see formatting gate the build job instead of running beside it, that is a one-line change to add `needs: format`.


.git-blame-ignore-revs
paladine, line 1: is this file required?

paladine (approved):

.git-blame-ignore-revs
mparrett: Not required, and not really part of this PR. It comes from #676, which this one is stacked on.

It lists the #661 formatting commit so `git blame` skips past it. GitHub applies the file automatically; locally it needs `git config blame.ignoreRevsFile .git-blame-ignore-revs`. Happy to drop it from #676 if you'd rather not carry it.

Also want to mention this PR is still a draft and its `format` job is red on its own branch. `dotnet format` doesn't agree with itself across SDK feature bands.  CI resolves `10.0.x` to 10.0.400, which wants different indentation on the `goto` label in `CPUCore.cs` than 10.0.101 does locally. So the gate isn't meaningful until the SDK is pinned, and pinning via `global.json` constrains every contributor's toolchain, which felt like it needed some discussion. I could add it here, send it separately, or drop. Let me know your thoughts!

mparrett: cc: @enusbaum — thoughts?

PR #679, opened 2026-08-22
Fix crash and corruption bugs in the host and strcpy paths

mparrett:
Split out of #653 so these can be reviewed independently of #662. #653 changes `BtrieveFileProcessor.cs`, which #662 rewrites; nothing here touches Btrieve. What remains on #653 is the SQLite hardening, parked until #662 lands.

## Why

These are crash and corruption fixes from a continuously running instance hosting MajorMUD (`WCCMMUD`) and Trade Wars 2002 (`HVSTW`) side by side with live telnet users — 108 crash reports between 2025-12-30 and 2026-02-28. Every item below except the `ReadOnlySpan` null check and the `TELL` stub crashed the emulator or corrupted module state on that instance, rather than turning up in a code audit.

## What changed

**`lonrou` ran on every module even when disabled.** `ProcessLONROU` set `session.OutputEnabled = ModuleDoLoginRoutine` and then invoked `lonrou` for every loaded module regardless: the setting silenced the output but did not gate the call. With Trade Wars loaded, its login routine runs `DosGetSegDesc` while silent, after which MajorMUD answers every command with "Your command had no effect." Skipping the call when the setting is off restores command dispatch, so `Module.DoLoginRoutine=False` now does what its name says. The shipped default is unchanged.

I have not isolated which piece of state the silent call clobbers, so this gates a known-bad interaction rather than fixing its root cause.

**`strcpy` crashed instead of truncating.** A module passing a destination into a segment that does not exist threw out of the exported function and killed the CPU loop. `SetArray` reaches the destination through a span, and a missing segment faults two ways depending on the offset: a non-zero offset throws `ArgumentOutOfRangeException` inside `AsSpan`, while offset 0 yields an empty span whose indexer throws `IndexOutOfRangeException`. Only the first is an `ArgumentException`, so the original handler missed the second. Both are now caught: copy what fits, NUL-terminate, log, and return `dest` in `DX:AX` as the real call does. The live crash came from a destination in segment `0x1001`, which was never allocated. Two in-segment overflows were also logged there, 42 bytes into 56 and 42 into 71, which are the too-short-destination case the original handler already truncated.

**`strcpy` checked the source buffer before the null guard.** `GetString` ran on the source pointer before `FarPtr.Empty` was tested, so a null source dereferenced first. The guard now runs first and sets the destination to an empty string.

**Dictionary mutated during iteration.** Nightly cleanup iterated `_modules` and removed from it inside the loop, throwing `InvalidOperationException`. It now snapshots before iterating, copies each module's cleanup commands before `Dispose`, and matches export keys on `"{id}-"` so `WCCMMUD` no longer also matches `WCCMMUD2`.

**`ReadOnlySpan<T>` compared against null.** `fieldSpec == null` in `FsdUtility` is always false on a value type, so the guard never ran. Replaced with the intended length check. The matching fix in `BtrieveFileProcessor` stays on #653, since #662 rewrites that file.

**`TELL` (598) stubbed.** Trade Wars calls it; it is an empty stub alongside the other ignored ordinals.

## Known remaining

`ProcessLONROU_FromRlogin` still calls every module's `lonrou` with output disabled, the same shape as the telnet bug above. I left it alone because I have no rlogin drop-in setup to verify a change against, and an unverified fix there seemed worse than a documented gap. I would rather take it as a separate PR, though I will fold it in here if you prefer.

## Verification

- A new test pins the unallocated-destination `strcpy` crash at both offsets, covering each exception path; both cases fail with the handler removed. A second covers the null-source ordering. The existing test covers the too-short-but-valid destination.
- Full suite green: 2602 passed, 0 failed, as of 2026-08-22.
- Conflict-free against every other open PR, including #662.
- Both modules run side by side through login and gameplay without the corruption or crashes above.

### DIFF
--- a/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
+++ b/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
@@ -607,6 +607,7 @@ namespace MBBSEmu.HostProcess.ExportedModules
                 case 513: //RSTWIN -- restore window parameters (local screen)
                 case 82: //BAUDAT -- Set color attribute based on baud, ignoring as ANSI is always on
                 case 549: //SHOCHL -- Show legend for channel with any attribute, ignoring
+                case 598: //TELL -- Send message to another user, stub for now
                     break;
                 case 73:
                     applyem();
--- a/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
+++ b/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
@@ -1631,7 +1631,50 @@ namespace MBBSEmu.HostProcess.ExportedModules
             }
             else
             {
-                Module.Memory.SetArray(destinationPointer, inputBuffer);
+                try
+                {
+                    Module.Memory.SetArray(destinationPointer, inputBuffer);
+                }
+                catch (Exception ex) when (ex is ArgumentException or IndexOutOfRangeException)
+                {
+                    // Real DOS strcpy() would overflow into adjacent memory.
+                    // In emulation, copy as much as possible in-segment to avoid crashing.
+                    var copied = 0;
+                    for (; copied < inputBuffer.Length; copied++)
+                    {
+                        var targetOffset = destinationPointer.Offset + copied;
+                        if (targetOffset > ushort.MaxValue)
+                            break;
+
+                        try
+                        {
+                            Module.Memory.SetByte(destinationPointer.Segment, (ushort)targetOffset, inputBuffer[copied]);
+                        }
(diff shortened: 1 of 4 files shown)


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
paladine: could move this up to 1622 as well

paladine (approved):

MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
mparrett: Done in e9c0ba11 — hoisted above the branch, so both assignments are gone.

paladine (approved):

PR #681, opened 2026-08-23
perf(cpu): remove duplicate Tick() decode, avoid FPU store allocations

paladine:
Tick() called Memory.GetInstruction()/GetCurrentOperationSize() twice per instruction unconditionally, in every build: the #if DEBUG block that looked like it gated the second call actually spanned past it, so the fetch+decode ran twice every tick. Removing the redundant call measured ~18% higher throughput on a MOV/CMP/JMP microbenchmark.

WriteToDestination(float/double), Op_Fistp, and Op_Fst used BitConverter.GetBytes(...) to write FPU results to memory, heap-allocating a new byte[] on every FPU store. Switched to stackalloc + TryWriteBytes, measured ~9-10% higher throughput on an FLD/FADD/FSTP microbenchmark.

### DIFF
--- a/MBBSEmu/CPU/CPUCore.cs
+++ b/MBBSEmu/CPU/CPUCore.cs
@@ -405,8 +405,6 @@ namespace MBBSEmu.CPU
             _currentInstructionPointer.Offset = Registers.IP;
             _currentInstructionPointer.Segment = Registers.CS;
 #endif
-            _currentInstruction = Memory.GetInstruction(Registers.CS, Registers.IP);
-            _currentOperationSize = GetCurrentOperationSize();
 
             InstructionCounter++;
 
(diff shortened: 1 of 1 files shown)


mparrett (approved):
Read through this — it looks correct to me, and the duplicate decode is real: master calls `Memory.GetInstruction` + `GetCurrentOperationSize` twice unconditionally in `Tick()`, once after each `#if DEBUG` block, with neither call inside one. Removing the second is right, and keeping the first is necessary since the debug logging reads `_currentInstruction.IP16`.

I checked each `stackalloc` size against its operand type, since a wrong size there truncates silently rather than throwing:

- `WriteToDestination(float)` → 4
- `WriteToDestination(double)` → 4 for the `(float)` cast, 8 for the double
- `Op_Fistp` → `valueToSave` is declared `int`, so 4

All six sites write the same bytes `BitConverter.GetBytes` did.

Three notes, none blocking:

1. There's a DEBUG-only behavior change worth a line in the description: `Debugger.Break()` now fires between the only decode and execution, so changing `CS:IP` at that breakpoint no longer gets picked up by a re-decode the way the second call allowed. Obscure, but it's the one semantic difference in the change.

2. `BitConverter.TryWriteBytes` returns a `bool` that's discarded at all six sites. Safe today since every span is exactly sized — just fragile if a size ever changes.

3. Not introduced here, but it's in the code you're touching: `FSTP m80` writes 8 bytes to a 10-byte `Float80` destination, and the matching read at `CPUCore.cs:1208` also reads 8. Master does the same, so this PR doesn't regress it. Happy to file it separately if it isn't already tracked.

PR #693, opened 2026-09-04
Fix formatted output buffer lifecycle

manicpop:
## Summary

- preserve `PRFBUF` and `PRFPTR` across repeated `outprf()` calls so one formatted message can be sent to multiple channels
- record that the current buffer has been sent, then start a fresh message on the next `prf()`, `prfmsg()`, or `prfmlt()` call
- keep the sent-state flag synchronized with existing buffer resets in valid-channel `SetState()`, `clrprf()`, and `injoth()`
- add regression tests for repeated sends, each formatting entry point, and channel-independent callbacks

## Background

PR #686 fixed one half of the MajorBBS formatted-output lifecycle: `outprf()` must leave the current buffer intact because modules can format a message once and call `outprf()` repeatedly to broadcast it to several channels.

The other half is that the first subsequent formatting operation starts a new message. Without that transition, `prf()`, `prfmsg()`, or `prfmlt()` appends new content to the message that was already sent. This can duplicate earlier output and can eventually produce malformed or truncated displays.

Clearing the buffer directly in `outprf()` would break broadcasts. This change instead tracks whether the current buffer has been sent:

1. Formatting calls append normally while the buffer has not been sent.
2. `outprf()` sends the buffer and marks it as sent without changing `PRFBUF` or `PRFPTR`.
3. Additional `outprf()` calls resend the same buffer.
4. The next formatting call resets `PRFPTR` to `PRFBUF`, terminates the old contents, clears the sent flag, and writes the new message.

## State ownership

`_prfBufferSent` is intentionally an instance field on `Majorbbs`:

- `MbbsHost.GetFunctions()` creates exported-module instances using a key containing the module identifier.
- Each loaded `MbbsModule` therefore receives its own `Majorbbs` instance.
- Each `MbbsModule` also owns its own emulated memory core containing its `PRFBUF` and `PRFPTR`.
- Duplicate module identifiers are rejected during configuration.

The flag and the buffer it describes consequently have the same per-module lifetime and cannot leak between modules.

## `SetState()` behavior

This revision retains current master's existing valid-channel `SetState()` behavior: it clears `PRFBUF` and resets `PRFPTR`, and now also clears `_prfBufferSent` to keep the managed state synchronized.

Unlike the original revision of this PR, it does not move that reset ahead of the `ushort.MaxValue` early return. Channel-independent callbacks preserve the module-local lifecycle state. If the previous buffer was sent, their first formatting call starts fresh through the same normal transition. This avoids broadening `SetState()` behavior solely to implement the output lifecycle.

`injoth()` already consumes and clears the formatted-output buffer, so it now explicitly clears `_prfBufferSent` as well.

## Testing

- focused `outprf` lifecycle tests: 6 passed
- complete MBBSEmu test suite: 2,737 passed, 0 failed
- self-contained Linux x64 publish succeeded
- runtime-tested with classic Galactic Empire, ge-next, and ge-arena loaded simultaneously
- all three modules loaded and operated correctly with no cross-module output contamination
- ge-next `SCAN RANGE` displays complete output without duplicated headers or truncation

The full suite was run with `DOTNET_USE_POLLING_FILE_WATCHER=1` because the test host otherwise exceeded its per-user inotify watcher limit. The initial failures from that run were fixture-construction errors caused by the host resource limit; with polling enabled, the suite passed completely.

### DIFF
--- a/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
+++ b/MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
@@ -108,6 +108,13 @@ namespace MBBSEmu.HostProcess.ExportedModules
         private int _sprIndex;
 
         /// <summary>
+        ///     True after outprf() sends the current buffer. Repeated outprf() calls
+        ///     resend it, while the next formatting call starts a new message. This
+        ///     state belongs to the same per-module Majorbbs instance as PRFBUF.
+        /// </summary>
+        private bool _prfBufferSent;
+
+        /// <summary>
         ///     Int 21h handler
         /// </summary>
         private readonly Int21h _int21h;
@@ -359,9 +366,10 @@ namespace MBBSEmu.HostProcess.ExportedModules
             //Reset NXTCMD
             Module.Memory.SetPointer("NXTCMD", Module.Memory.GetVariablePointer("INPUT"));
 
-            //Reset PRFPTR
+            //A valid-channel callback begins with an empty formatted-output buffer.
             Module.Memory.SetPointer("PRFPTR", Module.Memory.GetVariablePointer("PRFBUF"));
             Module.Memory.SetZero(Module.Memory.GetVariablePointer("PRFBUF"), 0x4000);
+            _prfBufferSent = false;
 
             //Set FSDSCB Pointer for Current User
             var channelFsdscb = Module.Memory.GetOrAllocateVariablePointer($"FSD-Fsdscb-{ChannelNumber}", FsdscbStruct.Size);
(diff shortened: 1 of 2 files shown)


MBBSEmu/HostProcess/ExportedModules/Majorbbs.cs
```diff
@@ -104,6 +104,12 @@ private FarPtr _currentMcvFile
         /// </summary>
         private int _sprIndex;
 
+        /// <summary>
+        ///     True after outprf() sends the current buffer. Repeated outprf() calls
+        ///     resend it, while the next formatting call starts a new message.
+        /// </summary>
+        private bool _prfBufferSent;
```
enusbaum, line 115: I believe Majorbbs.cs and its state are shared across modules, would this cause issues if the user was running more than 1 module at the same time? 

We may need to reset this value in SetState() as well to ensure we're not carrying over the `_prfBufferSent` from one module to another in the main event loop.

This will need to be verified though.

enusbaum (changes requested):

paladine (approved):

PR #698, opened 2026-09-04
Implement RSMODES cleanup state

manicpop:
## Summary

- export the MajorBBS `rsmodes` per-channel reset-mode array at ordinal 504
- initialize each channel to `NORMRS`
- set all channels to `NANSRS` when the nightly cleanup grace period begins
- set the modes again immediately before cleanup disconnects users

## Background

MajorBBS sets `rsmodes[]` to its configured non-normal reset mode when the cleanup grace period begins, then sets it again in `hupall()` before invoking module hangup routines. Modules use this to distinguish an ordinary lost-carrier hangup from a system cleanup.

Without `RSMODES`, modules can treat cleanup as an ordinary disconnect and apply gameplay consequences intended only for a real hangup.

Timer callbacks only signal the host worker loop; module memory is updated from the worker thread. All configured channels are updated, including vacant channels that may be assigned during the grace period.

## Testing

- focused `rsmodes_Tests` and existing `NightlyCleanup_Tests` pass
- verified `rsmodes[0]` reports `NORMRS` during normal operation
- verified it changes to `NANSRS` at the configured cleanup time, before warning messages begin
- verified a combat-locked ge-next ship survives forced nightly cleanup

### DIFF
--- a/MBBSEmu/HostProcess/MbbsHost.cs
+++ b/MBBSEmu/HostProcess/MbbsHost.cs
@@ -100,6 +100,11 @@ namespace MBBSEmu.HostProcess
         private readonly Timer _cleanupTimer;
 
         /// <summary>
+        ///     Timer that marks the beginning of the nightly cleanup grace period
+        /// </summary>
+        private readonly Timer _cleanupStartTimer;
+
+        /// <summary>
         ///     Timer that triggers when nightly cleanup warning messages should start
         /// </summary>
         private Timer _cleanupWarningTimer;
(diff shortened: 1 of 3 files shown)


enusbaum (approved):

MBBSEmu/HostProcess/MbbsHost.cs
```diff
@@ -1511,6 +1533,16 @@ private void DoNightlyCleanup()
             Start(moduleConfigurations);
         }
 
+        private void SetCleanupResetModes()
+        {
+            foreach (var module in _modules.Values)
+            {
+                if (module.ExportedModuleDictionary.TryGetValue(Majorbbs.Segment, out var exportedModule) &&
+                    exportedModule is Majorbbs majorbbs)
+                    majorbbs.SetResetModes(Majorbbs.NANSRS);
```
paladine: I don't see this getting reset back to NORMRS?

PR #701, opened 2026-09-05
Preserve non-ASCII bytes in strncpy

manicpop:
## Summary

- copy `strncpy` source data directly from emulated memory rather than decoding it as an ASCII .NET string
- preserve non-ASCII byte values such as `0xAD`
- add a regression test covering a non-ASCII source byte and the returned destination pointer

## Background

MajorBBS `strncpy` operates on raw bytes. MBBSEmu previously read the source through `GetParameterString()`, which decodes it with `Encoding.ASCII`. Bytes outside ASCII are replaced with `?` (`0x3F`) before being written to the destination.

Classic Galactic Empire exposes this difference. Its security subsystem uses a private record key beginning with byte `0xAD` and copies that byte with `strncpy` when filtering the record from the roster. Under MBBSEmu, the copied byte became `0x3F`, so the comparison failed and the security record appeared as a player. Copying from `GetParameterStringSpan()` preserves MajorBBS behavior and keeps the record out of the roster.

## Testing

- focused `strncpy_Tests`: 5 passed
- verified with classic Galactic Empire under MBBSEmu that the security record no longer appears in the roster

### DIFF
--- a/MBBSEmu.Tests/ExportedModules/Majorbbs/strncpy_Tests.cs
+++ b/MBBSEmu.Tests/ExportedModules/Majorbbs/strncpy_Tests.cs
@@ -45,5 +45,31 @@ namespace MBBSEmu.Tests.ExportedModules.Majorbbs
             Assert.Equal(destinationStringPointer.Segment, mbbsEmuCpuRegisters.DX);
             Assert.Equal(destinationStringPointer.Offset, mbbsEmuCpuRegisters.AX);
         }
+
+        [Fact]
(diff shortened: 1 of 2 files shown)


paladine (commented):
The more I look at this, the more I think GetParameterString is just wrong and should not be using Encoding.ASCII, for the very reasons this PR was created. So instead of just fixing strncpy, we should focus on fixing all places where this might be broken

@enusbaum interested in your thoughts here, we probably shouldn't be using the encoding routines in this method for the stated reason

paladine (approved):

